Issue #6, opened 2015-03-03
Global default timeout


strukturedkaos:
@poteto - First of all, thanks for creating this awesome add-on.

I was wondering if there is a way to set an app-wide timeout value, so I don't have to set it for every flash message?

Thanks!

poteto:
Hey, thanks! I'll be adding this soon

Issue #10, opened 2015-03-08
Add theme options to configuration


poteto:

poteto:
In favour of keeping this addon minimal and customizable, I've decided against including anything style related.

Issue #9, opened 2015-03-08
Service should be configurable globally


poteto:

johnotander:
I'm wondering if you already have an approach you're considering for handling the global configuration for the service.

My initial idea for an approach was just to expand upon the `defaultX` properties like so:

```js
defaultTimeout() {
  return this._getGlobalDefaults().timeout || 3000;
},

// ... 

_getGlobalDefaults() {
  return Ember.ENV.flashMessageDefaults || {};
}
```

This has the benefit of working with the existing API (which I know is used by the flash object itself). But, it's a bit verbose, too. Is this similar to what you were intending @poteto? Or were you intending on going a different route?

poteto:
I'm happy to rip out internal API for something better! I don't mind verbosity as long as it's understandable. Originally, I was going to do something like this, which is probably not the best idea:

```javascript
// pseudo
import config from '../../config/environment';

const defaults = {
  timeout      : 3000,
  priority     : 100,
  sticky       : false,
  showProgress : false,
  type         : 'info'
};

export default Ember.Service.extend({
  _setDefaults: on('init', function() {
    const { flashMessageDefaults: userDefaults } = config;

    Object.keys(defaults).map((key) => {
      const classifiedKey    = key.classify();
      const defaultKey       = `default${classifiedKey}`;
      const defaultValue     = defaults[key];
      const userDefaultValue = userDefaults[key];
      let valueToSet         = userDefaultValue ? userDefaultValue : defaultValue;

      set(this, defaultKey, valueToSet);
    });
  })
});
```

The idea is to still allow people to set those defaults directly on the service, e.g.

```javascript
get(this, 'flashes').set('defaultTimeout', 5000);
```

johnotander:
Yeah, my proposed approach would be problematic because it wouldn't allow people to set the default directly. Observing `init` is a great approach. Though, I'd imagine we can merge `defaults` and `userDefaults` to clean up the loop?

```js
export default Ember.Service.extend({

  // ...

  _getDefaults() {
   const serviceDefaults = {
      timeout      : 3000,
      priority     : 100,
      sticky       : false,
      showProgress : false,
      type         : 'info'
    };

    const { flashMessageDefaults: userDefaults } = config;
    return Ember.$.extend({}, userDefaults, serviceDefaults);
  },

  _setDefaults: on('init', function() {
    var defaults = this._getDefaults();

    Object.keys(defaults).map((key) => {
      const classifiedKey = key.classify();
[...]

PR #26, opened 2015-03-14
[WIP] Add initial theme option to the app config

johnotander:
Firstly, thanks for an awesome addon.

Related to #10 

I'm not quite sure what options you want to expose in the
application config for flash message, so I've currently just
exposed a `flashTheme` config option. This is what the `alertType`
will use to infer the class prefix. One can still explicitly
set the `messageType` in the component declaratively, I've added
a test to verify that as well.

Is this along the lines of what you were thinking? What else
would you like to see here?

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -5,11 +5,11 @@ const { computed, getWithDefault } = Ember;
 export default Ember.Component.extend({
   classNames        : [ 'flashMessage' ],
   classNameBindings : [ 'alertType' ],
-  messageStyle      : 'bootstrap',
+  messageStyle      : undefined,
 
   alertType: computed('flash.type', function() {
     const flashType    = getWithDefault(this, 'flash.type', '');
-    const messageStyle = getWithDefault(this, 'messageStyle', '');
+    const messageStyle = getWithDefault(this, 'messageStyle', Ember.ENV.flashTheme || 'bootstrap');
     let prefix         = 'alert alert-';
 
     if (messageStyle === 'foundation') {
--- a/tests/unit/components/flash-message-test.js
+++ b/tests/unit/components/flash-message-test.js
@@ -3,6 +3,7 @@ import {
   test
 } from 'ember-qunit';
 import FlashMessage from 'ember-cli-flash/flash/object';
+import Ember from 'ember';
 
 var flash;
 
@@ -18,6 +19,7 @@ moduleForComponent('flash-message', 'FlashMessageComponent', {
 
   afterEach() {
     flash = null;
+    Ember.ENV.flashTheme = null;
   }
 });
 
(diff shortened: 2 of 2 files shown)


poteto:
Hi John,

Thanks for being the first contributor! :)

I guess I probably should have wrote a better description and title for the issue. Basically I had planned to include a few themes (different CSS files) that you could import into your app. That said though, I do want to move all the default options in the service to an object in `config/environment.js` (or perhaps write our own option loader, I'm not sure yet), so this is a step in the right direction!

addon/components/flash-message.js
```diff
@@ -5,11 +5,11 @@ const { computed, getWithDefault } = Ember;
 export default Ember.Component.extend({
   classNames        : [ 'flashMessage' ],
   classNameBindings : [ 'alertType' ],
-  messageStyle      : 'bootstrap',
+  messageStyle      : undefined,
 
   alertType: computed('flash.type', function() {
     const flashType    = getWithDefault(this, 'flash.type', '');
-    const messageStyle = getWithDefault(this, 'messageStyle', '');
+    const messageStyle = getWithDefault(this, 'messageStyle', Ember.ENV.flashTheme || 'bootstrap');
```
poteto: Could we namespace this under a single object? Maybe something like... 

```javascript
flashMessageDefaults: {
  theme    : 'bootstrap',
  type     : 'foo',
  timeout  : 3000,
  priority : 100,
  sticky   : false
}
```

tests/unit/components/flash-message-test.js
```diff
@@ -43,6 +45,28 @@ test('#alertType returns the right type', function(assert) {
   assert.equal(component.get('alertType'), 'alert alert-test');
 });
 
+test('#alertType allows the messageStyle to be set', function(assert) {
```
poteto: This feels like a bit of a redundant test

addon/components/flash-message.js
poteto: Also actually, we should move this into the service instead. The flash message component and object don't need to know about defaults

johnotander:
Yay! : )

Ah, I see. When I get a chance tomorrow (or perhaps this evening) I will amend this PR to use a config object for the service defaults since that seems to be the way I'm heading with this right now. This should then resolve #6 instead of #10. I definitely misinterpreted your issue a bit, but if it can still be helpful, I'm glad!

tests/unit/components/flash-message-test.js
johnotander: Indeed. I think I got a little overzealous.

addon/components/flash-message.js
johnotander: Sounds good, can do.

poteto:
That'd be great, thanks! :)

johnotander:
Going to go ahead and close this to open a new PR targeted for #6, and #9.

poteto:
:+1:

Issue #18, opened 2015-03-09
Don't remove flash if mouseover-ed


poteto:

michalsnik:
Hi @poteto,
I see that this topic got quiet over a year ago.. Did it stop exactly where you closed #28, or were you doing something about it in a meanwhile?
This sounds like a must have feature to me, it's a bit annoying when it disappears with mouseover. I'm willing to help with this one in a spare time if you'd like.

sbatson5:
@michalsnik Have a PR for this and looking for reviews:
https://github.com/poteto/ember-cli-flash/pull/209

sbatson5:
@michalsnik would you be able to add the last commit directly from #209 to your project and see if it works properly? I've tested it on a few versions of Ember and it seems to work but want to get a practical test in place.

Let me know if you have any questions.

sbatson5:
Closed by #209

PR #28, opened 2015-03-16
prevent flash message destruction on mouseover

leojpod:
Thanks for the addon!
Here is an attempt to implement #18 

Not ready to be merged Yet! Unit tests and cie are missing!


leojpod:
I have some issue with the acceptance test writting. 
Somehow: `triggerEvent('selector', 'click');` is working while it does not look like `triggerEvent('selector', 'mouseEnter');` is. 
Any idea?

poteto:
I'll have a look! Thanks for contributing

tests/acceptance/integration-test.js
```diff
@@ -52,3 +52,42 @@ test('sticky messages are not removed automatically', function(assert) {
...
+  triggerEvent('.alert.alert-success', 'mouseenter');
+  
+  assert.ok(find('.alert.alert-success'), 'right after the rendering the flash message is there!');
+  andThen(() => {
+    assert.ok(find('.alert.alert-success'), 'the success message received the mouseenter and stayed on display');
+    assert.equal(find('.alert').length, 2); //there should be the sticky and the mouseover-ed
+    assert.equal(find('.alert.alert-success h6').text(), 'Success');
+    assert.equal(find('.alert.alert-success p').text(), 'Route transitioned successfully');
+  });
+  triggerEvent('.alert.alert-success', 'mouseleave');
```
poteto: Try moving this below `L84`. The issue is probably that the `triggerEvent` call happens synchronously before the `andThen` calls can finish

poteto:
As a side note, I wonder if there's a performant way of keeping track of how much time is left on the `run.later` call, so we can resume it instead of restarting it...

tests/acceptance/integration-test.js
leojpod: If I call `triggerEvent` within the `andThen` or after the flash messages will be already gone wouldn't they? 
I am not the best test coder so I shamelessly copied your test cases and tried to adjust them for the mouseover-ed feature. 
I even tried some ugly alert to make sure I would see if the component handler for `mouseenter` is called but no luck there. 
I'll keep looking!

leojpod:
About tracking the time: the solutions I found online consits mainly of using a custom method to setup the run.later and keep track of when it was fired and for how long which doesn't seem super efficient but why not?

I'm wondering if it is really something that should be implemented though: let's say there is a flash msg that interests me and I was a bit slow to react, the timeleft is now about 20 ms, I move my pointer over the message and in my hurry I go beyond the flash message, the timeleft has been set back to what was left of the 20 ms and now the popup is gone. 

Perhaps we should use a second delay for mouseover-ed flash messages, something like 250ms and whenever the mouseleft event is triggered call `destroyLater` with that value.

What do you think?

tests/acceptance/integration-test.js
leojpod: I found out what is happening there but I dont know how to work around it. I'll ask on discuss.ember.js

What is happening is that `triggerEvent` isn't called right after `visit('/');` but only once the flash messages disappeared (i.e. if I ask the flash message to stay up 1 sec the triggerEvent will be called a bit after that) kindof like if `triggerEvent` was wrapped up in `andThen`. I tried to put `visit` and `triggerEvent` in the same run loop but that doesn't seem to have much impact...

poteto:
Hey @leojpod, how's this going? Need any help getting it over the line?

leojpod:
Hey! 
Sorry I have been pretty busy lately... 
The feature seems to work but there is no test written for it yet. 
I wanted to try writing some by reopening the flash object and overriding the destroy method to call `assert.ok` instead and coupling it with `assert.expect` to make sure the correct number of tests are called but I haven't had time to get it done yet unfortunately. 
I'll try to get that done by the end of the week.

leojpod:
This is getting pretty frustrating ... Using Flash.reopen is working to test the "correctness" of the mouseover-ed feature but I can't figure out how to limit the "reopening" to one test... As it stands now, every most tests passing after the reopen will fail. Any idea?

tests/acceptance/integration-test.js
```diff
@@ -55,3 +58,21 @@ test('sticky messages are not removed automatically', function(assert) {
     assert.equal(find('.alert.alert-danger p').text(), 'You went offline');
   });
 });
+test('sticky messages are still removed when clicked', function(assert) {
+  assert.expect(5);
```
truenorth: Typically you don't need to `assert.expect` -- unless you're doing something async that the ember async helpers don't already handle. I'd remove this and see if the test still passes

PR #29, opened 2015-03-17
[WIP] Allow global service configuration

### LINKED ISSUE #6: Global default timeout

johnotander:
Here's the first pass, as discussed.

__TODO:__
- [x] Implement the service configuration from the ENV.
- [x] Use the config hook to set defaults.
- [x] A few tests to ensure defaults are being loaded from the config.
- [x] Document it.

Closes #6, partially address #9.

### DIFF
--- a/addon/services/flash-messages-service.js
+++ b/addon/services/flash-messages-service.js
@@ -5,6 +5,7 @@ const {
   computed,
   get: get,
   getWithDefault,
+  set,
   A: emberArray,
   on
 } = Ember;
@@ -12,14 +13,7 @@ const {
 export default Ember.Service.extend({
   queue               : emberArray([]),
   isEmpty             : computed.equal('queue.length', 0),
-
-  // refactor
-  defaultTimeout      : 3000,
-  defaultPriority     : 100,
-  defaultSticky       : false,
-  defaultShowProgress : false,
   defaultTypes        : [ 'success', 'info', 'warning', 'danger', 'alert', 'secondary' ],
-  defaultType         : 'info',
 
   arrangedQueue: computed.sort('queue', function(a, b) {
     if (a.priority < b.priority) {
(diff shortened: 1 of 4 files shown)


poteto:
Thanks! Do you have time to finish this in the next few days? Would be nice to have this in :)

johnotander:
Yeah, sorry for the delay. I will be able to get to it this evening.

addon/services/flash-messages-service.js
```diff
@@ -12,14 +13,7 @@ const {
 export default Ember.Service.extend({
   queue               : emberArray([]),
   isEmpty             : computed.equal('queue.length', 0),
-
-  // refactor
-  defaultTimeout      : 3000,
-  defaultPriority     : 100,
-  defaultSticky       : false,
-  defaultShowProgress : false,
   defaultTypes        : [ 'success', 'info', 'warning', 'danger', 'alert', 'secondary' ],
```
poteto: We could probably move defaultTypes into the global config as well. That way you could overwrite the thing if you didn't need any of those types. It would then make registering types manually with registerType redundant and we can make that a private method

johnotander: Yeah, that makes sense. Will do.

poteto:
@johnotander Sorry to be a nag, but I'm happy to take over the rest if you're busy with other things!

johnotander:
Not a nag at all, I'm sorry to leave an open WIP PR hanging for you.

I've got a bit of time right now to see if I can get this finished up. If I fail to do so, feel free to take it over :), as I'll then be pretty busy through the week.

johnotander:
I've got everything together and working now, however, I am unable to import the config to leverage in place of `Ember.ENV`. I'm not sure if it is because of the addon context or what, but `import config from '../config/environment'` errors out.

poteto:
@johnotander Thanks! If you look at @rwjblue's [ember-cli-content-security-policy](https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L4), you can see how he does it

johnotander:
Radical, thanks for the link. I will dive into it first thing tomorrow morning.

MichalBryxi:
This is a must-have thing. Do you need a help with finishing this?

johnotander:
Hey @MichalBryxi,

I've just pushed the rest of what I have so far, using the `config` hook to set defaults for flash messages. However, in the tests it seemed like the config wasn't being read in, so I had to keep the current `_getDefaults` code in the service (otherwise the tests bombed).

As a result, the default options are currently defined twice, which will need to be remedied. I will be pretty busy through this weekend, so if you want to pick up where I've left off, please feel free!

poteto:
@johnotander I checked out your branch and had a go at the dummy config for tests. Have a look at #36 and let me know what you think

johnotander:
Looks awesome @poteto! Thanks for taking over this work in progress. :smile_cat:

PR #34, opened 2015-03-27
allow passing arbitrary options to flash messages

cowboyd:
Sometimes an application wants to customize how its flashes are
rendered on a per-flash basis. This allows the application to set
arbitrary options onto the instance of `FlashMessage` so that display
can be customized.

The use case I'm ultimately going for is to embed arbitrary components onto the flash.

For example, in the OSX update message, there is a non-trivial UI
contained entirely inside the flash message, complete with buttons and
dropdowns.

magic_newsletter_-_february_2014

By allowing the caller of `flashes.add()` to pass any context data along
to the flash object, the template that ultimately renders the flash can
be as rich as it needs to be. For example:

```hbs
{{#each flashes.queue as |flash|}}
  {{#flash-message flash=flash as |cmp flash|}}
    {{#if flash.template}}
      {{render flash.template flash.context}}
    {{else}}
      <h6>{{c.flashType}}</h6>
      <p>{{flash.message}}</p>
    {{/if}}
  {{/flash-message}}
{{/each}}

```

then in app code

```js
flashes.success({
  template: "operation-successful-what-now",
  context: operationDetails
})
```

I wasn't quite sure whether to mixin the application options directly onto the flash message, or whether to keep them in their own separate `options`, or similarly named object. I made the first attempt directly on the flash object since writing `flash.options.template` and `flash.options.context` seemed awkward.


poteto:
Hey, thanks for this PR! :) :+1: 

I originally wrapped the flash component's template in an [`{{#if template}}` block](https://github.com/poteto/ember-cli-flash/blob/develop/app/templates/components/flash-message.hbs#L1) for a similar reason, although it's globally defined and you can't pass in a template on a per flash basis. Having arbitrary options is definitely an important use case, and it's in line with what we're trying to also do on #36 / #29. I like the simplicity of having it set on the flash object and I'm happy to merge this in if you're satisfied with it!

cowboyd:
I'm satisfied with it for the moment. If you'd like I can wait a week or so and see how the ergonomics feel inside a few applications and report back to you. At the same time though, I'm fine if you'd like to merge it now.

Also, thanks for a great library... such a timesaver!

poteto:
Thanks! I'll merge after the merge conflicts are sorted out, unless you'd like me to do it

cowboyd:
All set!

poteto:
Awesome, thanks! :D

PR #41, opened 2015-03-31
s/Boostrap/Bootstrap

bsclifton:


poteto:
Typing is hard :P Thanks Brian! :)

bsclifton:
NP- I was giggling when I popped out at me, like its a framework for your Boo or perhaps a Halloween themed framework :P

PR #46, opened 2015-04-04
Minor build/test tweaks.

rwjblue:
* Remove ember-data from build.  This library does not use ember-data, and occasionally having ember-data in the build causes test failures (ember-data 1.0.0-beta.16.1 does not work with Ember canary at this point).
* Test across many versions of Ember.
* Use ember-cli-babel instead of custom preprocessor.
* Fix style warning on 1.12.0-beta.1+.


poteto:
Awesome stuff, thanks Rob!

PR #59, opened 2015-04-22
Fix typo in readme

truenorth:
This could be potentially confusing :)


poteto:
:+1: Good catch! Typing is hard :p

PR #67, opened 2015-05-22
Added extendTimeout as a configuration option to make animation possible

### LINKED ISSUE #52: Transition hooks for animating out
Great addon!

I've just included a CSS keyframe animation to fade in the flash messages but just scratching my head as to how to fade them out again.

The addon seems to remove the messages instantly from the DOM, is there any way it could remove the active class first, then waiting for the JS transition event before removing it?

If you have approach that works already it would be fantastic to hear it.

Thanks

jrjohnson:
### what works
My solution to fix #52

Setting disappearTimeout on a message (or in the config) will cause the class
‘disappearing’ to be added to the message after the timeout.  This
should allow for CSS animation to take place before the message element
is destroyed like:

```scss
.alert {
  opacity: 1;
  transition: all 2s linear;

  &.disappearing {
    opacity: 0;
  }
}
```
### what needs work
I kicked around the names for the option *disappearTimeout* and the class *disappearing* with a few people around the office and this is the best we could come up with.  Other ideas would be welcome. 

The way you are using services, objects, and components here is foreign territory for me.  So there are areas where I :monkey: copied for sure.  I got some great help on the ember slack channel from *cody.mcculloch* and *davidpett*.  Thanks to them if any of this is right.

I wrote an acceptance test in tests/acceptance/integration-test.js but I wasn't able to get it running, you can see it is commented out.  For some reason my message was always marked as sticky so the timeouts never get set.  Any idea why that would happen?  Maybe unit tests cover the case sufficiently, your call.

Lastly my editor did some white space cleanup in README.  If you liked it the way it was I can put it back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,12 +66,13 @@ You can also pass in options to custom messages:
 
 ```javascript
 Ember.get(this, 'flashMessages').add({
-  message      : 'I like alpacas',
-  type         : 'alpaca'
-  timeout      : 500,
-  priority     : 200,
-  sticky       : true,
-  showProgress : true
+  message            : 'I like alpacas',
+  type               : 'alpaca'
+  timeout            : 500,
+  priority           : 200,
+  sticky             : true,
+  showProgress       : true,
+  extendedTimeout    : 500,
 });
 
 Ember.get(this, 'flashMessages').success('This is amazing', {
@@ -115,6 +116,13 @@ Ember.get(this, 'flashMessages').success('This is amazing', {
   Default: `false`
 
   To show a progress bar in the flash message, set this to true.
+
+- `extendedTimeout?: number`
+
+  Default: `0`
+
+  Number of milliseconds before a flash message is removed to add the class 'exiting' to the element.  This can be used to animate the removal of messages with a transition.
+
 
 ### Arbitrary options
 You can also add arbitrary options to messages:
--- a/addon/flash/object.js
+++ b/addon/flash/object.js
@@ -8,6 +8,7 @@ const {
   getWithDefault,
   run,
   on,
+  setProperties,
   Evented
 } = Ember;
 
@@ -19,8 +20,11 @@ export default Ember.Object.extend(Evented, {
   isErrorType: computed.equal('type', 'error').readOnly(),
 
   defaultTimeout: computed.readOnly('flashService.defaultTimeout'),
+  defaultExtendedTimeout: computed.readOnly('flashService.defaultExtendedTimeout'),
   queue: computed.readOnly('flashService.queue'),
   timer: null,
+  exitingTimer: null,
+  exiting: false,
 
   destroyMessage() {
     const queue = get(this, 'queue');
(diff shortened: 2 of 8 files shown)


poteto:
Hey, thanks for opening this PR, I'll review it shortly!

jrjohnson:
Any thoughts or changes to this approach?  I will try and get the integration test running, but I wanted to be sure this overall was a good idea before I dug into it any further.

poteto:
Hey, sorry about the delay. I'm reviewing this right now.

README.md
```diff
@@ -66,12 +66,13 @@ You can also pass in options to custom messages:
...
-  priority     : 200,
-  sticky       : true,
-  showProgress : true
+  message            : 'I like alpacas',
+  type               : 'alpaca'
+  timeout            : 500,
+  priority           : 200,
+  sticky             : true,
+  showProgress       : true,
+  disappearTimeout : 500,
```
poteto: So with the README changes, I'd prefer if you'd revert the whitespace changes, and update the formatting of L#75 to match the lines above it (aligned by `:`). As for naming, `disappearTimeout` makes sense, but maybe `extendedTimeout` would be more correct?

README.md
```diff
@@ -99,23 +100,29 @@ Ember.get(this, 'flashMessages').success('This is amazing', {
...
 - `showProgress?: boolean`
-  
+
   Default: `false`
 
   To show a progress bar in the flash message, set this to true.
 
+- `disappearTimeout?: number`
+
+  Default: `false`
```
poteto: Incorrect default value

addon/flash/object.js
```diff
@@ -50,6 +51,22 @@ export default Ember.Object.extend(Evented, {
     }
   },
 
+  disappearMessage() {
+    this._super();
```
poteto: Unnecessary super call

addon/flash/object.js
```diff
@@ -50,6 +51,22 @@ export default Ember.Object.extend(Evented, {
...
+
+    const timer = get(this, 'disappearTimer');
+    const isDestroying = get(this, 'isDestroying');
+    const isDestroyed = get(this, 'isDestroyed');
+
+    if (!isDestroyed && !isDestroying) {
+      set(this, 'disappearing', true);
```
poteto: Nitpick: this could be:

```js
setProperties(this, {
  disappearing: true,
  disappearTimer: null
});
```

config/environment.js
```diff
@@ -7,6 +7,7 @@ module.exports = function(/* environment, appConfig */) {
       priority           : 100,
       sticky             : false,
       showProgress       : false,
+      disappearTimeout   : false,
```
poteto: Should be `0`

tests/acceptance/integration-test.js
```diff
@@ -48,3 +48,22 @@ test('high priority messages are rendered on top', function(assert) {
     assert.equal(find('.alert p').first().text(), 'It is going to rain tomorrow');
   });
 });
+
```
poteto: I wouldn't worry about the integration test, you don't have to add one here.

tests/dummy/config/environment.js
```diff
@@ -22,6 +22,7 @@ module.exports = function(environment) {
       priority           : 100,
       sticky             : true,
       showProgress       : false,
+      disappearTimeout   : false,
```
poteto: Should be `0`

poteto:
Thanks for this PR @jrjohnson! Sorry about the delay, last week was kind of intense for me. It looks OK, just a few minor nitpicks and a thing to fix.

I'm not 100% certain about `disappearTimeout` though. I think `extendedTimeout` would be more correct.

addon/flash/object.js
```diff
@@ -61,5 +78,13 @@ export default Ember.Object.extend(Evented, {
     const destroyTimer = run.later(this, 'destroyMessage', timeout);
 
     set(this, 'timer', destroyTimer);
+
+    const defaultDisapearTimeout = get(this, 'disappearTimeout');
```
poteto: `disappearMessage` will never run, because the flash object would have been destroyed by the time it gets there. You should move the logic into `destroyMessage` instead, above [L33](https://github.com/jrjohnson/ember-cli-flash/blob/animateremoval/addon/flash/object.js#L33) so that it can delay destruction

jrjohnson:
No worries on the delay thanks for the thorough review.

My thought was that you would set disappearTimeout to a length of time less than timeout.  So 
```
dissapearTimeout: 1000
timeout: 3000
```
Then the class would get added after one second and the object would get destroyed after 3.  That does requires the developer to figure this out.  

I think from the notes that you are thinking extendTimeout would work in conjunction with timeout.  So I would set:
```
extendTimeout: 1000
timeout: 3000
```
Then after 3 second the class would be added and one second later the message would be destroyed.

Am I reading that correctly, if so I will make the changes and write some new tests.

poteto:
Yup, exactly! And by default it's `0`, so it still works as before.

jrjohnson:
OK - extendTimeout is now the option which ended up being a much cleaner approach.  I also rebased to cleanup.

poteto:
Looks good! Let me take a closer look tomorrow morning and I'll merge it in

PR #73, opened 2015-06-24
Ember 2.0 update (not more EnumerableUtils)

stefanpenner:


raytiley:
you need to do `map.call` and `forEach.call` when used this way.

stefanpenner:
actual PR coming

PR #78, opened 2015-06-27
Add active class to flash messages upon DOM insertion

truenorth:
In order to make the CSS animation work properly, some style has to change after inserting the flash message into the DOM. Setting active to true doesn't have the desired effect, so (I think) progress bars have never worked.

### DIFF
--- a/tests/dummy/config/environment.js
+++ b/tests/dummy/config/environment.js
@@ -17,6 +17,13 @@ module.exports = function(environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
+    contentSecurityPolicy: {
+      'img-src': '*',
+      'connect-src': '*',
+      'font-src': '*',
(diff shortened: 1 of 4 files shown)


tests/dummy/config/environment.js
```diff
@@ -17,6 +17,13 @@ module.exports = function(environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
+    contentSecurityPolicy: {
+      'img-src': '*',
+      'connect-src': '*',
+      'font-src': '*',
+      'script-src': '\'unsafe-inline\' *',
+      'style-src': '\'unsafe-inline\' *'
+    },
```
truenorth: Unrelated, but this gets rid of `[Report Only]` errors in the dummy app

poteto:
:+1: Thanks Mike!

PR #86, opened 2015-07-12
Restore ember 1.11.x compatibility

truenorth:
:tada:

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "devDependencies": {
     "broccoli-asset-rev": "^2.0.2",
     "ember-cli": "1.13.1",
-    "ember-cli-app-version": "0.4.0",
+    "ember-cli-app-version": "0.3.5",
     "ember-cli-content-security-policy": "0.4.0",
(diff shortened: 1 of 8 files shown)


poteto:
Thanks @truenorth, this is awesome! I didn't know there was a polyfill for it.

package.json
```diff
@@ -24,7 +24,7 @@
   "devDependencies": {
     "broccoli-asset-rev": "^2.0.2",
     "ember-cli": "1.13.1",
-    "ember-cli-app-version": "0.4.0",
+    "ember-cli-app-version": "0.3.5",
```
poteto: Why was this downgraded?

truenorth:
Your co-worker wrote it :)

package.json
truenorth: `ember-cli-app-version: 0.4.0` makes use of [instanceInitializers](http://emberjs.com/blog/2015/05/13/ember-1-12-released.html#toc_instance-initializers) which were introduced in `1.12.0`.

poteto: :+1:

PR #87, opened 2015-07-13
Allow css transition on message creation

mjanda:
I haven't found a way to use css transition to animate message appearance, so this should allow to put transitions on .active class.

If there is some existing way I've missed, let me know. Thank you

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -65,7 +65,7 @@ export default Ember.Component.extend({
   }),
 
   _setActive: on('didInsertElement', function() {
-    run.scheduleOnce('afterRender', this, () => {
+    run.next(() => {
       set(this, 'active', true);
     });
   }),


poteto:
You should already be able to add animations with the `active` class

jrm2k6:
Do you have an example with it working when targeting the ```active``` class.
It won't fade or translate in our case. It seems that there is no point where the DOM element for the flash message is not having the ```active``` class.

sdhull:
I agree that an example in the readme where animations are actually added would be useful.

As I've been playing with it, I've found the following scss works pretty well (I'll open a PR to add it to the readme):
```scss
.alert {
  opacity: 0;
  position: relative;
  left: 100px;

  transition: all 700ms cubic-bezier(0.68, -0.55, 0.265, 1.55);

  &.active {
    opacity: 1;
    left: 0px;

    &.exiting {
      opacity: 0;
      left: 100px;
    }
  }
}
```
I recommend easings.net as a great resource for choosing your own easing function.

One thing to note: the exiting animation is broken if you have `sticky: true`

Issue #92, opened 2015-08-09
Uncaught TypeError: ember_new_computed.default.readOnly is not a function


pedrokost:
Using ember-cli-flash@1.3.3
Ember      : 1.13.6
Ember Data : 2.0.0-beta.1
jQuery     : 2.1.4

in line 14 of `object.js`:

```javascript
queue: ember_new_computed['default'].readOnly('flashService.queue'),
```

pedrokost:
This "bug" was introduced in version `1.3.2`. All previous versions work fine (between `1.1.4` and `1.3.1`).

pedrokost:
This issue is holding my upgrade to ember 2.0. Have others been able to upgrade ember with the latest version of the addon?

grantovich:
I also see this error when attempting to use ember-cli-flash 1.3.4 with Ember 1.13.9. Tested with the not-yet-released 1.3.5 and still got the same error.

poteto:
Thanks for the report and super sorry for not following up on this earlier. I'm looking into it, hopefully @rwjblue can also chime in

rwjblue:
@poteto - Do we have a reproduction?

grantovich:
I'm personally unable to reproduce the issue on a fresh-generated app using Ember-CLI 1.13.8, so it may be some interaction with the other dependencies of my app (and @pedrokost's). I haven't had time yet to investigate further, but I can confirm that downgrading to 1.3.1 also resolves the issue for me.

rwjblue:
What is the output of `npm ls ember-new-computed`?

rwjblue:
@poteto - I suspect the issue is that some other addon is including ember-new-computed@1.0.0 that is overriding the version that you are using here.  1.0.0 did not provide `computed.foo` macros, which would explain this issue.

Work around solution for that would be to use `import computed from 'ember-new-computed'; const { computed: { readOnly } } = Ember;` for the other ones, but I'd like to confirm the output of `npm ls ember-new-computed` above and get whatever is locked to 1.0.0 fixed.

grantovich:
@rwjblue Thanks for the pointer! It appears I was using an older version of `ember-leaflet` that was locked to 1.0.0. The current version no longer uses `ember-new-computed` at all, so I should probably upgrade and the issue will be resolved.

[...]

PR #109, opened 2015-10-16
Update ember-new-computed to allow a bit more flexibility.

rwjblue:
This does not fix the issue reported in https://github.com/poteto/ember-cli-flash/issues/92, but if all consumers of ember-new-computed used SemVer allowed drifting versions we should be able to avoid issues of different deps clobbering others.


poteto:
Thanks Rob!

PR #113, opened 2015-10-28
Fix warn deprecation and bad cp

XrXr:
Fixes two issues:

- Ember 2.1 deprecation warnings for calling `Ember.warn()` without an options hash with id
- Computed properties that sets itself to the object they are on, after calling `Ember.set()` with any value on them.

For detail please read my commit messages.
Cherry picking 9bb1113 from my repo and running it against develop would fail.

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -10,7 +10,7 @@ const {
   run,
   on,
   get,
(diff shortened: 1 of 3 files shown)


poteto:
Closed by #134

PR #115, opened 2015-11-03
Starts the timeout on message render instead of creation

pzuraq:
@poteto Changes the behavior so that messages will start timeout when they are rendered rather than when they are created, allowing for sequencing and more fine tuned control by users.

Wasn't sure if this would constitute a major or minor version bump! Could argue that it may break older behavior, but I find that doubtful unless people were using `flash` objects without the `flash-message` component. Let me know and I'll bump `package.json` accordingly.

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -11,6 +11,7 @@ const {
   on,
   get,
   set,
+  observer
 } = Ember;
 const {
   readOnly,
@@ -58,6 +59,14 @@ export default Component.extend({
       return this;
     }
   }),
+
+  // Observer used here for backwards compatibility with Ember 1.11/1.12, replace
+  // with on('didRender') once possible.
+  _startTimers: on('didInsertElement', observer('flash', function() {
+    run.schedule('afterRender', this, () => {
+      get(this, 'flash').startTimer();
+    });
+  })),
 
   _setActive: on('didInsertElement', function() {
     run.scheduleOnce('afterRender', this, () => {
--- a/tests/index.html
+++ b/tests/index.html
@@ -24,7 +24,7 @@
     <script src="assets/vendor.js"></script>
     <script src="assets/test-support.js"></script>
     <script src="assets/dummy.js"></script>
-    <script src="testem.js"></script>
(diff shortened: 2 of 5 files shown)


addon/components/flash-message.js
```diff
@@ -59,6 +60,14 @@ export default Component.extend({
     }
   }),
 
+  // Observer used here for backwards compatibility with Ember 1.11/1.12, replace
+  // with on('didRender') once possible.
+  _startTimers: on('didInsertElement', observer('flash', function() {
```
poteto: Why does this need to be an observer? It only needs to fire one time when inserted into the DOM.

tests/index.html
```diff
@@ -24,7 +24,7 @@
     <script src="assets/vendor.js"></script>
     <script src="assets/test-support.js"></script>
     <script src="assets/dummy.js"></script>
-    <script src="testem.js"></script>
+    <script src="testem.js" integrity="" ></script>
```
poteto: :+1:

addon/components/flash-message.js
pzuraq: In testing I found that the timers wouldn't start with just `didInsertElement` in some cases. I'm pretty sure this happens if you change the `flash` field on the component, for instance if the component renders the first message, it times out, and then it renders the second message without actually destroying the component.

Once 1.11 and 1.12 are dropped this can be solved with using `didRender`.

poteto: That's odd. I'll investigate and let you know, I'm very against using observers unless strictly necessary

pzuraq: Agreed, definitely don't want to use them unless there is no other way.

Here's the specific example that required this:

```
{{#flash-message flash=flashMessages.arrangedQueue.firstObject}}
  ...
{{/flash-message}}
```

poteto:
Thanks!

pzuraq:
@poteto Any word on this? I can rebase if the solution looks good, the only reason we had to use observer was for the Ember 1.10 compatibility I believe. We could use `didRender` if we dropped backwards compat instead.

PR #121, opened 2015-11-25
Add active class in run.next instead of afterRender queue

kmiyashiro:
Currently, the active class is added at the exact same time the message is
rendered (despite the afterRender queue). This will ensure that the active
class is added 1ms after it is inserted into the DOM, thus triggering any CSS
transitions from naked to active states.


poteto:
Thanks!

PR #132, opened 2016-02-16
Allow baseline config to cascade through addons

ksnyde:


poteto:
This approach wouldn't work as ember-cli-flash can no longer import the consuming app's environment. I'm updating the addon to 2.0 soon and will address this issue there, so closing this for now.

PR #135, opened 2016-02-17
Update ember-cli

poteto:
- Removed unnecessary readOnly sets on certain CPs
- Updated use of `warn` to include `id`


poteto:
@homu r+

PR #136, opened 2016-02-17
Cascading Addons, take 2

ksnyde:
Not sure what the timeline is for 2.0.0 but this should address your concern about end-user config. I'll use it as a fork for now if 2.x is close to close but if you expect a delay then maybe this is worthy?

Thanks for such a great addon. Looking forward to 2.x.


poteto:
Thanks, I have a PR already open addressing it though – https://github.com/poteto/ember-cli-flash/pull/134

poteto:
Closed by #134

PR #148, opened 2016-02-23
Fix calling set on destroyed object error

mitchlloyd:
I have a login route that renders flash messages in a certain part of the template. Then the primary parent route of the application renders flash messages again. I was getting a "cannot call set on destroyed object" from [this code](https://github.com/poteto/ember-cli-flash/blob/1.3.12/addon/components/flash-message.js#L53).

I added a component integration test to show the issue: after a component is torn down, `_setActive` still tries to change a property on that component. 

The simple fix is to check for `isDestroyed` on that component, but probably there is a more wholistic refactor needed.

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -15,6 +15,10 @@ const {
   readOnly,
   bool
 } = computed;
+const {
+  next,
+  cancel
+} = run;
 
 export default Component.extend({
   layout,
@@ -49,9 +53,10 @@ export default Component.extend({
   }),
 
   _setActive: on('didInsertElement', function() {
-    run.next(this, () => {
+    const pendingSet = next(this, () => {
       set(this, 'active', true);
     });
+    set(this, 'pendingSet', pendingSet);
   }),
 
   progressDuration: computed('flash.showProgress', {
@@ -73,6 +78,7 @@ export default Component.extend({
   willDestroy() {
     this._super();
     this._destroyFlashMessage();
+    cancel(get(this, 'pendingSet'));
   },
 
   // private
--- a/tests/integration/components/flash-message-test.js
+++ b/tests/integration/components/flash-message-test.js
@@ -1,0 +1,36 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import FlashMessage from 'ember-cli-flash/flash/object';
+
+moduleForComponent('flash-message', 'Integration | Component | flash message', {
+  integration: true
+});
+
+test('it renders a flash message', function(assert) {
+  this.set('flash', FlashMessage.create({ message: 'hi', sticky: true }));
+
+  this.render(hbs`
+    {{#flash-message flash=flash as |component flash|}}
+      {{flash.message}}
+    {{/flash-message}}
+  `);
+
+  assert.equal(this.$().text().trim(), 'hi');
+});
+
+test('it does not error when quickly removed from the DOM', function(assert) {
+  this.set('flash', FlashMessage.create({ message: 'hi', sticky: true }));
+  this.set('flag', true);
+
+  this.render(hbs`
+    {{#if flag}}
+      {{#flash-message flash=flash as |component flash|}}
+        {{flash.message}}
+      {{/flash-message}}
+    {{/if}}
+  `);
+
+  this.set('flag', false);
+
+  assert.ok(this.get('flash').isDestroyed, 'Flash Object isDestroyed');
+});


mitchlloyd:
After some more poking around I realized what happened in my case:

1. I rendered a list of flash messages from `flashMessages.queue`.
2. Inside of an action, I added a new flash message. This caused the rendering of a new `flash-message` component to begin.
3. I immediately `transitionTo` a new route, tearing down the component and destroying the FlashMessage.
4. The deferred `set` operation is called on the destroyed component, throwing one of our favorite Ember errors :smile:

In this case I needed to understand that every time a flash message is rendered it is tied to the lifecycle of that component. If I had added the flash message after the transition was complete it would have worked. I still think this may be a valuable test to integrate to avoid this error and just be consistent with housekeeping of cleaning up timers.

Overall this destroy behavior was surprising to me. If I move between routes that each render their own flash messages, they will disappear, but if flash messages are rendered in the application route (for instance) they will never go away until dismissed or cleared. Perhaps this is a better pattern than what I was expecting, but my first thought was that I would need to clear the messages myself unless the user dismissed them or there was some timeout set for them.

In any event, I hope something in there is helpful. Thanks for the great library!

ahx:
:+1: 
I am having the same issue when running acceptance tests for one app. I am using the flash message to show a global "saving"… "done" message and got this error ("set on destroyed object") when calling .clearMessages() between "saving" and "done" messages. This PR fixes that.

addon/components/flash-message.js
```diff
@@ -49,7 +49,7 @@ export default Component.extend({
   }),
 
   _setActive: on('didInsertElement', function() {
-    run.next(this, () => {
+    this.pendingSet = run.next(this, () => {
```
poteto: Can we use `set(this, 'pendingSet', ...)` instead?

addon/components/flash-message.js
```diff
@@ -73,6 +73,7 @@ export default Component.extend({
   willDestroy() {
     this._super();
     this._destroyFlashMessage();
+    run.cancel(this.pendingSet);
```
poteto: Can we destructure `cancel`, then use `cancel(get(this, 'pendingSet'))` instead?

tests/integration/components/flash-message-test.js
```diff
@@ -0,0 +1,36 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import FlashMessage from 'ember-cli-flash/flash/object';
+
+moduleForComponent('flash-message', 'Integration | Component | flash message', {
+  integration: true
+});
+
+test('it renders a flash messages', function(assert) {
```
poteto: `s/messages/message`

poteto:
Thanks for the PR! Just a few minor comments.

addon/components/flash-message.js
mitchlloyd: Yep! When all laid out, it looks a little tougher to read to me though. When nothing should observe an assignment I often forego `set`. But I also understand sticking with the established style in the project if always using `set` is more consistent.  Have a look and let me know if you want to go back to using the `=` assignment:

```javascript
  _setActive: on('didInsertElement', function() {
    set(this, 'pendingSet', run.next(this, () => {
      set(this, 'active', true);
    }));
  })
```

addon/components/flash-message.js
mitchlloyd: Sure. I'll destructure `run.next` as well for consistency.

Again here using `get` and `set` is a little misleading to my eyes since it makes me expect that there is some observation occurring. I'll make the change and let me know if you want to switch it back.

mitchlloyd:
@poteto Cool! I just updated the commit. I'm interested to know what you think about using `get`/`set` vs `=` for assignment.

addon/components/flash-message.js
```diff
@@ -49,9 +53,9 @@ export default Component.extend({
   }),
 
   _setActive: on('didInsertElement', function() {
-    run.next(this, () => {
+    set(this, 'pendingSet', next(this, () => {
```
poteto: You could do:

```js
const pendingSet = next(this, ...);
```

tests/integration/components/flash-message-test.js
mitchlloyd: :+1:

poteto:
Generally speaking I prefer to use `get/set` as a convention, and I only use vanilla setters in the `init` hook. When I'm dealing with a POJO though, I may opt to use vanilla setters unless something is observing my POJO

addon/components/flash-message.js
mitchlloyd: Just modified and pushed again.

poteto:
@homu r+

PR #149, opened 2016-03-03
Compute guid on .toString() version of the message.

louim:
Hello!

This PR allows using Ember i18n service, or any string which is an Ember SafeString and have the `{ preventDuplicates: true }` work by computing the guid value on the stringified version of the message. Before, it computed the guid directly, so two different instance of the same objet would get different guid. I added a test to cover the case.

Note: there is a test failing, but it was failing before too.


poteto:
LGTM, thanks!

@homu r+

PR #155, opened 2016-04-12
Fix Ember.merge -> Ember.assign transition and run tests on 2.4 and 2.5

martndemus:
See: http://emberjs.com/blog/2016/04/11/ember-2-5-released.html#toc_code-ember-assign-code


poteto:
@homu r+

PR #159, opened 2016-04-21
Remove obsolete contribution doc

jrowlingson:
It appears that recent contributions are no longer following the gitflow branching model.


poteto:
@homu r+

PR #157, opened 2016-04-21
Add destroyOnClick option

jrowlingson:
This PR adds the following option:

- `destroyOnClick?: number`

  Default: `true`

  By default, flash messages will be destroyed on click.  Disabling this can be useful if the message supports user interaction.

PR #163, opened 2016-05-08
Disable auto click

abhilashlr:
@poteto this is an updated version of the closing the flash message on click.


jrowlingson:
I believe this has been addressed by https://github.com/poteto/ember-cli-flash/pull/157

PR #187, opened 2016-08-04
Allow flash-message to not have any default class as prefix

cibernox:
With this change, if the component receives a `messageStyle="one-of-my-own"` it doesn't
generate it just generates an "alert-{{TYPE}}" (instead of "alert alert-{{type}}"), since
that is the styles expected by bootstrap.

The default behavior is still the same, this just allows the user to opt out.

This allows me by example to create my own component that extends from the default one and provide different classes (since in my project the "alert" class is used for another kind of component, and I want to use "flash" instead.

``` js
import FlashMessageComponent from 'ember-cli-flash/components/flash-message';

export default FlashMessageComponent.extend({
  classNames: ['flash'],
  messageStyle: 'custom'
});

```

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -34,9 +34,10 @@ export default Component.extend({
     get() {
       const flashType = getWithDefault(this, 'flash.type', '');
       const messageStyle = getWithDefault(this, 'messageStyle', '');
-      let prefix = 'alert alert-';
-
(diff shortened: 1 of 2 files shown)


sbatson5:
@cibernox You good if I close this? 😅

cibernox:
@sbatson5 be my guest 👍

PR #204, opened 2016-09-14
Remove unnecesary ember-new-computed addon

cibernox:
If this addons is 1.12+ (and I think that it safe to assume it is), this
addon is not needed anymore.

### DIFF
--- a/addon/flash/object.js
+++ b/addon/flash/object.js
@@ -1,17 +1,14 @@
 import Ember from 'ember';
 import customComputed from '../utils/computed';
-import computed from 'ember-new-computed';
 
 const {
   Object: EmberObject,
+  computed: { readOnly },
+  run: { later, cancel },
   Evented,
   get,
-  run: { later, cancel },
   set
 } = Ember;
-const {
-  readOnly
-} = computed;
 
 export default EmberObject.extend(Evented, {
   timer: null,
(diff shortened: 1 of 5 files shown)


cibernox:
Rebased

addon/flash/object.js
```diff
@@ -1,13 +1,13 @@
...
-import computed from 'ember-new-computed';
 
 const {
   Object: EmberObject,
   run: { later, cancel },
   Evented,
   get,
-  set
+  set,
+  computed
```
sbatson5: Can we also add `readOnly` like we do with `later` and `cancel`?

```
...
set,
computed,
computed: { readOnly }
```

cibernox: Done.

sbatson5: Ah, it looks we still need to pull in `computed` as well as `computed: { readOnly }`

cibernox: I don't think so. It's not used within that file. This timeout seems unrelated.

cibernox: I'll push again with no changes

sbatson5: Ah, my mistake. Sorry about that.

cibernox: It's green now.

cibernox:
Rebased again.

sbatson5:
This LGTM 👍 . Sorry for the delay in merging!

PR #209, opened 2016-10-02
[#18] - defer removal of flash message

sbatson5:
- [x] Prevent flash message removal on mouse over per https://github.com/poteto/ember-cli-flash/issues/18
- [x] Resume exit timers with remaining time per https://github.com/poteto/ember-cli-flash/pull/28#issuecomment-81991231
- [x] Tests
  - [x] Add unit test for deferring removal
  - [x] Add integration test for mouseover events


tests/unit/flash/object-test.js
```diff
@@ -1,14 +1,14 @@
...
 import FlashMessage from 'ember-cli-flash/flash/object';
+import sinon from 'sinon';
 
 const testTimerDuration = 50;
 const {
   run,
   get
 } = Ember;
 let flash = null;
-let SANDBOX = {};
```
sbatson5: Didn't look like we ever used this.

tests/integration/components/flash-message-test.js
```diff
@@ -34,3 +40,72 @@ test('it does not error when quickly removed from the DOM', function(assert) {
...
+  assert.equal(this.$().text().trim(), 'hi');
+  assert.notOk(destroyMessage.calledOnce, 'flash has not been destroyed yet');
+
+  clock.tick(timeoutDefault);
+  assert.ok(destroyMessage.calledOnce, 'flash is destroyed after timeout');
+
+  clock.restore();
+});
+
+test('flash message is removed after timeout', function(assert) {
```
sbatson5: 🤔 This test seems to fail on Ember 1.13 and 2.0. Curious if it is a `sinon` failure

poteto: Try running `ember try:one ember-1.13 --- ember test --serve` and see if there are any errors in the console

sbatson5: @poteto thoughts on checking the version of Ember for this test? While debugging, it seems that my spy gets overwritten with the original method on Ember 1.13. It almost looks like the object gets re-created.  I ran this commit on an app running 1.13 and can confirm it stays when moused over--it just doesn't seem to handle the spy very well.

sbatson5:
@poteto Do you think this in a good place to merge? I'd love to close out that issue.

poteto:
LGTM

PR #210, opened 2016-10-10
Add base class 'flash-message' to flash component

### LINKED ISSUE #133: add base css class to {{flash-message}}
I'm not using bootstrap or foundation. It would be nice to have a more specific way of selecting than: `<div id="ember1749" class="ember-view alert alert-info active">` so it doesn't collide with other potential selectors. I propose "flash-message" so it would be `<div id="ember1749" class="ember-view flash-message alert alert-info active">`. I would be open to doing the PR if it's approved.

jfelchner:
Closes #133

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -24,6 +24,7 @@ export default Component.extend({
   layout,
   active: false,
   messageStyle: 'bootstrap',
+  classNames: 'flash-message',
   classNameBindings: ['alertType', 'active', 'exiting'],
 
   showProgressBar: readOnly('flash.showProgress'),


addon/components/flash-message.js
```diff
@@ -24,6 +24,7 @@ export default Component.extend({
   layout,
   active: false,
   messageStyle: 'bootstrap',
+  classNames: 'flash-message',
```
sbatson5: I believe that `classNames` still needs to be an array even if we are only adding 1 class.
`classNames: ['flash-message'],`

sbatson5:
@jfelchner Thoughts on adding a quick integration test to assert that the class is present?

sbatson5:
going to merge this and add an integration test as a separate PR.

Issue #154, opened 2016-04-11
Access to flash object on .add() (conflicts with "Fluent API")


davidgovea:
First off, thank you for this project - it's been very helpful!

I noticed a change in behavior while upgrading from a far-outdated version:

It looks like the 1.3.7 release ("Fluent API") makes a pattern of returning `this` for chainability.

Previously, `flashMessages.add({ ... })` would return a Flash object. We are defining some dynamic computed properties on this flash object, as well as listening for `didDestroyMessage`.

I submit that obtaining the flash message object is an important feature. Chainability is nice, but I haven't run into pain regarding multiple calls on the service. Maybe what I'm doing is not advisable - just wanted to provide feedback on our use-case. For the time being, I've simply overridden the `add()` method with the old style (returning `this._enqueue(..flashmessage..)`) - addons make it really easy to do so.

Thanks for reading!

davidgovea:
I've been thinking about my approach, and perhaps it's less-than-Embery.
I'm implementing this library into a second project, and I'm going to rely more heavily on components this time (like readme example `{{component flash.componentName}}` rather than dynamically augmenting the flash object)

ahx:
I also have another use-case for accessing the flash object. I am using a custom type, "progress" where I am showing a progress-bar inside the flash message. Currently I am clearing and re-adding the flash-message, but having access to the flash object would be nicer. I am not interested in the chaining aspect.

aroberts:
I also depend on this capability - I need to be able to update the message being displayed. I'll be taking the first approach (overriding `add()` in the service) until flash message object is exposed somewhere via the api.

sbatson5:
@ahx or @aroberts Curious how you would want this used. Would something as simple as an `addAndReturnFlash` method that is not chainable work? I personally haven't had a use-case for accessing the flash object directly

ahx:
@sbatson5 Yes, I think would work for my use-case. Maybe call it `create` instead? Do you know about a use-case for chaining?

sbatson5:
I think `create` is a bit too vague and would easily get used over `add` in instances where people would want to chain. I like being verbose in the case of something like `addAndReturn`. There's probably a better name, but it is escaping me.

sbatson5:
@ahx or @aroberts let me know what you think of: https://github.com/poteto/ember-cli-flash/pull/214

This allows us to keep the chaining but you can still return the flash object:
[...]

PR #214, opened 2016-10-31
add chainable method for returning flash object

sbatson5:
## What's in this PR
This should address the issue raised by #154.  A fluent API allows for chaining methods, but does not give access to the flash object. For users who want to manipulate the object directly, they can access it with the `getFlashObject` method, which will return the last object added to the queue.

- [x] Add chainable method that returns the flash object
- [x] Add test coverage in service test
- [x] Update Readme to include documentation on `getFlashObject`

### DIFF
--- a/addon/services/flash-messages.js
+++ b/addon/services/flash-messages.js
@@ -68,6 +68,21 @@ export default Service.extend({
     return this;
   },
 
+  peekFirst() {
+    return get(this, 'queue.firstObject');
+  },
+
+  peekLast() {
+    return get(this, 'queue.lastObject');
+  },
+
+  getFlashObject() {
+    const errorText = 'A flass message must be added before it can be returned';
+    assert(errorText, get(this, 'queue').length);
+
+    return this.peekLast();
+  },
+
   _newFlashMessage(options = {}) {
     assert('The flash message cannot be empty.', options.message);
 
--- a/README.md
+++ b/README.md
@@ -177,11 +177,23 @@ This makes use of the [component helper](http://emberjs.com/blog/2015/03/27/embe
 ```
 
 ### Clearing all messages on screen
-It's best practise to use flash messages sparingly, only when you need to notify the user of something. If you're sending too many messages, and need a way for your users to clear all messages from screen, you can use this method:
+It's best practice to use flash messages sparingly, only when you need to notify the user of something. If you're sending too many messages, and need a way for your users to clear all messages from screen, you can use this method:
 
 ```javascript
 Ember.get(this, 'flashMessages').clearMessages();
 ```
+
+### Returning flash object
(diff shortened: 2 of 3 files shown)


addon/services/flash-messages.js
```diff
@@ -68,6 +68,13 @@ export default Service.extend({
     return this;
   },
 
+  getFlashObject() {
```
sbatson5: @poteto Thoughts on if this is acceptable based on the existing pattern? Addresses #154 but it isn't quite as fluent.  Also, thoughts on renaming to `getLastFlashObject` since it is a public method and could be called anywhere?

panthony: IMO `getLastFlashObject` holds more meaning than `getFlashObject`.

Or since we are talking about a queue of message, why not provide accessors such as `peekFirst` and `peekLast` ?

sbatson5: @panthony trying to wrap this up today. I like `peekLast` and `peekFirst` but what about a specific index?  Or should we not worry about that?

I kind of like `getFlashObject` because it implies that it is chainable and returns the flash object we just added.  If someone wants to get a specific index, they can access the queue from the service directly.

panthony: Here's my current implem:

https://github.com/cogniteev/ember-cli-flash/commit/21727b0149732300ec29cbd29434a08c38570804

sbatson5: Awesome!

I'm thinking of maybe having all 3 methods.  I like the readability of: 
`flash.add({ message: 'foo' }).getFlashObject();` as it returns the object we just added. `peekLast` doesn't make sense to chain.

Then we can still have `peekLast` and `peekFirst` but encourage that they aren't chainable but are public calls to our service.

Thoughts?

sbatson5:
@panthony let me know what you think of https://github.com/poteto/ember-cli-flash/pull/214/commits/a395f2cab2b9b5a755863d36fe6da2cca1cf71a7

panthony:
@sbatson5 wow I was not aware of the `firstObject` and `lastObject` accessor. That's pretty convenient. Never too late to learn something.

I'd say that maybe getFlashObject is redundant but it's a matter of taste I guess. Not sure why `peekXXX` would be 'less chainable'.

And didn't you say that `getLastFlashObject` was better ?

Or maybe rename `peekXXX` by `get{First|Last}FlashObject` so that you have the context of what you get no matter where you call the method ?

sbatson5:
@panthony I'm just thinking of the naming convention: `flashMessageService.add(object).getFlashObject` seems readable to me as it returns the object I _just_ added.

`flashMessageService.add(object).peekLast` seems like I am adding an object and then looking at the lost for some reason. The name doesn't imply that we are returning the result of our add. So in essence, `getFlashObject` is just an alias of `peekLast` but I think it is a bit more idiomatic when we are chaining.

panthony:
@sbatson5 Ok ! So this is looking great in that case :)

PR #218, opened 2016-11-29
remove flash.destroy as it fails on latest builds of Ember

sbatson5:
Master was getting failures when trying to run `flash.destroy()` on unit level tests of our component:
`Assertion Failed: Cannot call writeDeps after the object is destroyed.`

### DIFF
--- a/tests/unit/components/flash-message-test.js
+++ b/tests/unit/components/flash-message-test.js
@@ -26,10 +26,6 @@ moduleForComponent('flash-message', 'FlashMessageComponent', {
   },
 
   afterEach() {
-    run(() => {
-      flash.destroy();
-    });
-
     flash = null;
   }
 });


tests/unit/components/flash-message-test.js
```diff
@@ -26,10 +26,6 @@ moduleForComponent('flash-message', 'FlashMessageComponent', {
   },
 
   afterEach() {
-    run(() => {
-      flash.destroy();
```
sbatson5: @poteto Let me know if this fix seems appropriate. It resolves all the failing tests in beta and alpha:
https://circleci.com/gh/poteto/ember-cli-flash/323?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #226, opened 2017-01-02
Add a demo page ready to deploy to GH pages

### LINKED ISSUE #45: Build demo site

WillEngler:
WIP to close #45.

Important changes:

- Dummy app is clickable demo
- Uses [ember-cli-github-pages](https://github.com/poetic/ember-cli-github-pages) addon to make building and deploying dummy app to gh-pages branch straightforward with `ember github-pages:commit --message "gh-pages branch commit message"`. See example at [willengler.com/ember-cli-flash/](http://willengler.com/ember-cli-flash/)
- Points integration tests to new /test route so that you can edit the index route to your heart's content without worrying about messing up the integration tests
- Adds Bootstrap button styles to dummy app CSS

Blocking issue:

If you check out my example example (ha) you'll notice that the messages never disappear. When I do a fresh clone, install, and ember serve of `poteto/ember-cli-flash` I see similar behavior on the existing test app. The existing dummy app does not explicitly set a timeout, so the timeout should be 3000 ms. But when I run the original dummy app, the messages hang around indefinitely. **Do others see this behavior on a fresh install?** It seems implausible that I'd be the first to notice a bug preventing timeouts from working, but I don't know where to start looking for the problem in the dummy app.

### DIFF
--- a/tests/acceptance/main-test.js
+++ b/tests/acceptance/main-test.js
@@ -8,7 +8,7 @@ moduleForAcceptance('Acceptance | main');
 
 test('flash messages are rendered', function(assert) {
   assert.expect(7);
-  visit('/');
+  visit('/test');
 
   andThen(() => {
     assert.ok(find('.alert.alert-success'));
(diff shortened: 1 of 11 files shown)


WillEngler:
Just now seeing 49d2d593ed7a431420f8632f057c0c681ec023bc. So the groundwork had been laid for this a while back.

PR #230, opened 2017-05-26
Pass 'close' as closure action

caseywatts:
This adds in the ability to close the flash message in your own custom components. I tweaked the example in the readme to include a close button in the template there. I've also tried this in my own app :)

If we like this idea, we should probably make a separate example that has close on click set to false (since currently this example it wouldn't quite make sense why we want the close button when you can click on the whole thing to close it).

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,7 +384,7 @@ moduleFor('route:foo', 'Unit | Route | foo', {
 ```
 
 ## Styling
-This addon is minimal and does not currently ship with a stylesheet. You can style flash messages by targetting the appropriate alert class (Foundation or Bootstrap) in your CSS.
+This addon is minimal and does not currently ship with a stylesheet. You can style flash messages by targeting the appropriate alert class (Foundation or Bootstrap) in your CSS.
 
 ## License
 [MIT](LICENSE.md)
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ It also accepts your own template:
 ```
 
 ### Custom `close` action
-The `close` action is always passed to the component whether it is used or not. It can be used to implement your own close button, such an an `x` in the top-right corner.
+The `close` action is always passed to the component whether it is used or not. It can be used to implement your own close button, such as an `x` in the top-right corner.
 
 When using a custom `close` action, you will want to set `destroyOnClick=false` to override the default (`destroyOnClick=true`). You could do this globally in `flashMessageDefaults`.
 
--- a/README.md
+++ b/README.md
@@ -174,10 +174,11 @@ This makes use of the [component helper](http://emberjs.com/blog/2015/03/27/embe
 
 ```handlebars
 {{#each flashMessages.queue as |flash|}}
-  {{#flash-message flash=flash as |component flash|}}
+  {{#flash-message flash=flash as |component flash close|}}
     {{#if flash.componentName}}
       {{component flash.componentName content=flash.content}}
     {{else}}
+      <span {{action close}}>x</span>
       <h6>{{component.flashType}}</h6>
       <p>{{flash.message}}</p>
     {{/if}}
--- a/README.md
+++ b/README.md
@@ -174,11 +174,10 @@ This makes use of the [component helper](http://emberjs.com/blog/2015/03/27/embe
 
 ```handlebars
 {{#each flashMessages.queue as |flash|}}
-  {{#flash-message flash=flash as |component flash close|}}
+  {{#flash-message flash=flash as |component flash|}}
     {{#if flash.componentName}}
       {{component flash.componentName content=flash.content}}
     {{else}}
-      <a href="#" {{action close}}>close</a>
       <h6>{{component.flashType}}</h6>
       <p>{{flash.message}}</p>
     {{/if}}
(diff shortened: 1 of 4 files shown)


caseywatts:
I'm not 100% sure I want to yield it as another ordered parameter, but I don't quite wanna make a separate actions hash for it either lol
For this concern, I'm mostly worried about future edits to this addon - it works great as-is, no problem ✨

caseywatts:
I tried writing some tests for this feature, but they all suck so far lol:
https://github.com/poteto/ember-cli-flash/pull/231

caseywatts:
This branch has an integration test now - this PR should be good to go! :)

caseywatts:
@sbatson5 could you look over this when you get a chance? :)

sbatson5:
@caseywatts sorry for the delay. The test looks good. Just want to pull it into a dummy app real quick and test it. Should be able to get to that tonight

caseywatts:
sweet! thanks @sbatson5 :)

addon/templates/components/flash-message.hbs
```diff
@@ -1,5 +1,5 @@
 {{#if hasBlock}}
-  {{yield this flash}}
+  {{yield this flash (action 'close')}}
```
sbatson5: For consistency, we try to use double quotes on templates. Can you wrap `close` in double quotes?

tests/integration/components/flash-message-test.js
```diff
@@ -111,4 +111,29 @@ if (parseFloat(Ember.VERSION) > 2.0) {
...
   });
+
+  test('a custom component can use the close closure action', function(assert) {
+    assert.expect(3);
+
+    let destroyMessage = sinon.spy();
+    this.set('flash', FlashMessage.create({
+      message: 'flash message content',
+      sticky: true,
+      destroyOnClick: false,
```
sbatson5: 🤔 I'm wondering if there is a way for us to default this to false when yielding/using the close action. It could be confusing if both are enabled.

caseywatts: uh oh right, destroyOnClick=true is the default :/
I was surprised that was true when I first read the docs for this. And I'm surprised again just now haha (I wonder if other users are surprised by this default too?)

Hmm - we could at least document it in the README.md next to an example that uses `close`?

caseywatts:
I imagine this is a pretty common use case and I'd love to make it super straightforward, too.

-- 

This is a pretty common fork-in-the-road when designing Ember components. I'm not 100% that my suggestion is the better of the two paths, but I can provide more arguments for it at least :)

## contextual-component vs pass-in-more-arguments
@sbatson5 we're also going to want to put some classes on this - that'll lead us to three arguments.  Next (and this is probably going too far lol): some people may want this to be a styled `button`. If we wanted all of these, the API for this becomes something like this:

```
get(this, 'flashMessages').success('Success!!', {
  flashAction: {
    action: this.get('close'),
    label: 'Close',
    classNames: 'xyz',
    tagName: 'button'
  }
});
```

I think a [contextual component](https://www.mutuallyhuman.com/blog/2016/09/23/creating-an-accordion-with-contextual-components-in-ember) sort of thing gets around this sort of argument-passing-bloat pretty nicely :)

Also - the last argument `close` is optional to write, so I don't believe it'll break any existing code or examples. It should be a ~backwards-compatible change.

sbatson5:
@caseywatts ok, you convinced me :). Would you mind just updating the README and just mention to disable `destroyOnClick` if you want to use the `close` action? Once that's in, we're good to merge 👍 
(sorry for the delay)

addon/components/flash-message.js
```diff
@@ -107,5 +107,11 @@ export default Component.extend({
     if (flash) {
       flash.destroyMessage();
     }
+  },
+
+  actions: {
+    close() {
+      this._destroyFlashMessage();
```
sbatson5: What do you think of adding a warning via `Ember.Logger.warn` if `flash.destroyOnClick` is true?  Something like:

```
close() {
  if (get(this, 'flash.destroyOnClick')) {
    Logger.warn('This flash message has `destroyOnClick` enabled -- be sure to disable it when using the close action')
  }
  this._destroyFlashMessage();
}
```

bcardarella: @sbatson5 keep in mind that `warn` is stripped from Production builds: https://emberjs.com/api/#method_warn

caseywatts: I like the idea of this!
If it's stripped when this package is packaged (production build) then this one would be stripped out, hrm :/

How do other addons add warnings for users of the addon?

caseywatts:
updated the readme! 🎉

README.md
```diff
@@ -273,6 +272,20 @@ It also accepts your own template:
 {{/each}}
 ```
 
+### Custom `close` action
+The `close` action is always passed to the component whether it is used or not. It can be used to implement your own close button, such an an `x` in the top-right corner.
```
sbatson5: `such *as* an x`.

caseywatts:
typo corrected ✨

caseywatts:
and it's in! 🎉 

now that [master displays this option](https://github.com/poteto/ember-cli-flash#custom-close-action), we'll probably want to do a release soon so it matches up? cc @sbatson5 @Dhaulagiri

sbatson5:
Unfortunately, I don't have npm access to this package, so we will need @Dhaulagiri to publish and update for us.

Dhaulagiri:
published in 1.4.3 🎉

PR #231, opened 2017-05-26
Pass close as closure action - tests

caseywatts:
(I don't intend for this branch to be merged on its own, I just wanted it to be visible)

I took some stabs at testing for https://github.com/poteto/ember-cli-flash/pull/230 - but I don't like any of this so far lol

### DIFF
--- a/tests/integration/components/flash-message-test.js
+++ b/tests/integration/components/flash-message-test.js
@@ -3,6 +3,7 @@ import hbs from 'htmlbars-inline-precompile';
 import FlashMessage from 'ember-cli-flash/flash/object';
 import sinon from 'sinon';
 import Ember from 'ember';
+import wait from 'ember-test-helpers/wait';
 
 const timeoutDefault = 1000;
 const halfTimeout = 500;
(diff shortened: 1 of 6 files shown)


caseywatts:
ooh - I guess we could make a separate route in the dummy app to test this sort of component?

Is the dummy app demo page (in `tests/`) displayed anywhere for people to reference- or was it just intended to be that at some point in the past?

tests/integration/components/flash-message-test.js
```diff
@@ -111,4 +112,24 @@ if (parseFloat(Ember.VERSION) > 2.0) {
...
+      {{#flash-message flash=flash as |component flash close|}}
+        <a href="#" {{action close}}>close</a>
+        <a href="#">stay open</a>
+      {{/flash-message}}
+    `);
+    // this.$(":contains(stay open)").click();
+
+    this.$(":contains(close)").click();
+    return wait().then(() => {
+      assert.equal(this.$().text().trim(), '');
```
sbatson5: Hey @caseywatts. Thanks for the PR!

Rather than testing that the flash message itself disappears, you can stub the `destroyMessage` call on the flash-object with sinon, like we do here https://github.com/caseywatts/ember-cli-flash/blob/master/tests/integration/components/flash-message-test.js#L50

Then, you can test that it is `calledOnce` here like we do here:
https://github.com/caseywatts/ember-cli-flash/blob/master/tests/integration/components/flash-message-test.js#L69

Then we aren't testing that the flash message disappears, but that clicking `close` communicates with the flash-objects API in the way we expect.

Then maybe we can test the logic of it no longer rendering in an acceptance test 😄

caseywatts: sweet! thanks @sbatson5 :)
Just made an integration test and put it in that PR's main branch >:)

caseywatts:
got a test into the main PR so I'll close this side PR now :)

PR #234, opened 2017-07-03
Update to ember-cli 2.13

Dhaulagiri:
Built on top of https://github.com/poteto/ember-cli-flash/pull/233

* Swap out jshint for eslint
* Adds 2.8 lts to testing matrix (we should probably trim this entire list down in a separate PR)


Dhaulagiri:
I'll let #236 land and then bring back some of the smaller pieces of this PR afteward

PR #236, opened 2017-07-28
Upgraded to ember 2.14.1

bl4ckm0r3:
slightly changed tests to use wait

@poteto i had to change the sinon test, the component was actually calling the callback but sinon wasn't recognizing it (maybe the timer wasn't working?)


Dhaulagiri:
> i had to change the sinon test

I _think_ this may be fixed in Ember 2.15.  We should be able to confirm once that gets released soon.

bl4ckm0r3:
@Dhaulagiri i'll wait until 2.15 is out and we'll think about a strategy (probably close this and open a new pr?)

Dhaulagiri:
@bl4ckm0r3 I don't have a strong opinion.  The main thing I want to do is get tests passing on master again and then we can probably use most of what is in this PR and follow-on with a separate PR to update to ember-cli/ember 2.15 specifically.  If ember 2.15 does fix the reason our tests are failing we will get that "fix" for free since ember try will automatically run tests against it.

bower.json
```diff
@@ -1,8 +1,6 @@
 {
   "name": "ember-cli-flash",
   "dependencies": {
-    "ember": "~2.7.0",
-    "ember-cli-shims": "0.1.1",
     "ember-qunit-notifications": "0.1.0",
     "sinonjs": "^1.17.1"
```
Dhaulagiri: i'm pretty sure this can be removed as ember-sinon doesn't need this as of 0.5.1

bower.json
```diff
@@ -1,8 +1,6 @@
 {
   "name": "ember-cli-flash",
   "dependencies": {
-    "ember": "~2.7.0",
-    "ember-cli-shims": "0.1.1",
     "ember-qunit-notifications": "0.1.0",
```
Dhaulagiri: this can also be removed once ember-cli-qunit gets updated as it is in this PR.  We can do a separate PR to remove all the bower cruft.

.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
Dhaulagiri: I think the tabs in this file should only be 2 characters, not 2

bower.json
sbatson5: We should be able to move away from `bower` entirely with a few updates.

Dhaulagiri:
It doesn't look like ember 2.15 has changed anything with regard to the test failure, so something like what we have here seems necessary although I admit I'm not 100% familiar with the code to confidently give it a 👍 .

sbatson5:
@Dhaulagiri the intent of the test was to show that hovering over a message paused the dismissal timer (i.e. we could hover, wait an amount of time and test that our message was still there).

As @cmcclure pointed out in https://github.com/poteto/ember-cli-flash/issues/237, the problem/solution may have been over engineered (that's on me).  If we go the route of not having the timer paused, but instead just keep the message from destroying until mouse-leave, we should be able to go with these test updates.

bl4ckm0r3:
sorry for the delay i took some days off!
Have updated the pr, let me know what you think!
thanks

sbatson5:
@bl4ckm0r3 going to merge this in and PR another couple updates to fix the broken tests (larger issue unrelated to your updates)

Issue #242, opened 2017-09-18
Remove Sinon


sbatson5:
Since https://github.com/poteto/ember-cli-flash/issues/237 was resolved, we don't have a strong need for `sinon` as we don't need the fake timers.  The other instances where we use it (stubbing methods), we should have a better solution.

PR incoming.

PR #243, opened 2017-09-18
upgrade to 2.15

sbatson5:
## What's in this PR?

- Update to ember 2.15
- Remove some unused dependencies
- Move away from `sinon` for stubbing methods
- Update testem and remove jshint completely

closes: https://github.com/poteto/ember-cli-flash/issues/242

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "test": "tests"
   },
   "scripts": {
-    "start": "ember server",
     "build": "ember build",
+    "start": "ember server",
     "test": "ember try:each"
   },
   "repository": "https://github.com/poteto/ember-cli-flash",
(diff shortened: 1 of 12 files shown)


sbatson5:
@Dhaulagiri Think I could get a review? This PR should fix our failing tests and we are back to be green on every version in our `ember try`

tests/integration/components/flash-message-test.js
```diff
@@ -40,92 +41,77 @@ test('it does not error when quickly removed from the DOM', function(assert) {
...
-    this.$('#testFlash').mouseleave();
-        
-    return wait().then(() => {
-      assert.ok(
-        destroyMessage.calledOnce,
-        'flash waits remaining time from original timeout'
-      );
-    });
-  });
+  later(() => {
```
sbatson5: Rather than trying to be clever and use `sinon` to control the time, we'll just force it with `later`. It makes the test take a second longer to run, but we can get rid of `sinon` and this better matches the actual behavior.

package.json
```diff
@@ -25,32 +25,31 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "bower": "^1.8.2",
```
Dhaulagiri: i think this should go in devdeps since this is only to get ember try working, right?

package.json
```diff
@@ -25,32 +25,31 @@
...
     "ember-cli-babel": "^6.3.0",
     "ember-cli-htmlbars": "^2.0.1",
-    "ember-runtime-enumerable-includes-polyfill": "^1.0.1"
+    "ember-runtime-enumerable-includes-polyfill": "^2.0.0"
   },
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.5",
-    "ember-ajax": "3.0.0",
-    "ember-cli": "^2.14.1",
+    "ember-ajax": "^3.0.0",
```
Dhaulagiri: i think we could actually remove ember-ajax entirely here? it's part of the default blueprint but I don't think we are using it are we?

sbatson5: Nope, good call.

package.json
sbatson5: Good call. Fixed

PR #240, opened 2017-08-31
extendTimeout on sticky message

sbatson5:
Closes: https://github.com/poteto/ember-cli-flash/issues/228

## What's in this PR?

This PR adds the `exiting` class to messages that are sticky but have an extended timeout. This will honor exiting animations for messages that are clicked to dismiss.

Sticky message being dismissed with styling applied to `exiting` class:
exiting-message

PR #248, opened 2017-11-16
Use FlashObject#exitMessage on click or close

sdhull:
This allows developers' animations that leverage `.exiting` class to still work when clicking to dismiss messages.

### DIFF
--- a/addon/components/flash-message.js
+++ b/addon/components/flash-message.js
@@ -77,7 +77,7 @@ export default Component.extend({
     const destroyOnClick = getWithDefault(this, 'flash.destroyOnClick', true);
 
     if (destroyOnClick) {
-      this._destroyFlashMessage();
+      this._exitFlashMessage();
     }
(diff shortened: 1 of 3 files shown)


sbatson5:
Thanks so much for the PR @sdhull.  One thing to consider though is that the progress bar will still show even if the message is exiting--this could mess up expected animations.

Example:
animation-exit

Any easy way to fix this would be to conditionally show the progress bar if the `showProgress` boolean is present `and` if the message is not exiting.  You could update the `showProgressBar` computed to be:
```
showProgress: readOnly('flash.showProgress'),
notExiting: Ember.computed.not('flash.exiting'),
showProgressBar: Ember.computed.and('showProgress', 'notExiting'),
```
(But destructuring `computed` as with other CP's).  Let me know what you think.

sdhull:
Closing in favor of #240

PR #254, opened 2017-11-27
extendTimeout on sticky message

Dhaulagiri:


Dhaulagiri:
@sbatson5 I ended up reverting #240 because when reviewing I overlooked the fact that adding ember-concurrency required that dependency to be included in `dependencies` and not `devDependencies`.  When I pulled 1.6.0 into our app it broke looking for ember-concurrency.  Given this new information I'm wondering if there is a way to accomplish what you are doing here without ember-currency?  Requiring consumer of ember-cli-flash to include it if they aren't already using it (as we are not) quadruples the size of this addon which is something I'm hesitant to do.

sbatson5:
Ah, that's why I didn't see it when we pulled it into our existing app (we already had concurrency).

We can find another solution with the `run` loop, it will just require more cleanup when destroying the component.

sbatson5:
@Dhaulagiri Updated this PR to use the run loop instead of concurrent tasks. I've tested it a few times and everything seems to be working as expected.  Would you be able to pull it into a project and double check?

package.json
```diff
@@ -50,6 +50,7 @@
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.0.0",
+    "ember-maybe-import-regenerator": "^0.1.6",
```
Dhaulagiri: is this only used for tests?

Dhaulagiri:
Worked great in our app 👏

package.json
sbatson5: Yes. Was getting it for free with `ember-concurrency` apparently. Shouldn't add to the library size as it's just a `devDependency` but I can roll the tests back to nested `andThen`'s if that's preferred.

Dhaulagiri: Nope, that sounds great to me.  Just wanted to double check.

Dhaulagiri:
:shipit:

sbatson5:
Merged! Think we could cut another release?

Dhaulagiri:
Released in 1.6.2

PR #260, opened 2018-02-04
remove requirement for providing a message

### LINKED ISSUE #259: remove requirement for providing a message
We are happily using ember-cli-flash to display some custom notifications. Instead of just displaying a custom string, we sometimes have the need to render a component with additional functionality. This works pretty well, as we just need to pass the components name and then are able render it similar to what is shown in the docs.
```
this.get('flashMessages').add({
  component: 'group-undo-notification',
  model: group,
  message: '<none>',
  type: 'info'
});
```
However, this leads to situations where there simply is no message to display, as all the content is available through the component and additional objects that are passed in. This conflicts with the requirement to supply a message and we would like to propose to remove that requirement, as this just imposes unnecessary limitations to an otherwise very flexible addon.

st-h:
fixes #259 

I figured it would be reasonable to keep the `message` property required when `preventDuplicates` is enabled as that feature relies on the message string.

Please let me know or feel free to modify the added test, if you know of any better way to assert a thrown error than that:
```
assert.throws((() => {
  service.add({ });
}).message, 'Assertion Failed: The flash message cannot be empty when preventDuplicates is enabled.');
```

### DIFF
--- a/tests/unit/services/flash-messages-test.js
+++ b/tests/unit/services/flash-messages-test.js
@@ -169,7 +169,7 @@ test('when preventDuplicates is `false` setting a message is not required', func
     customProperty: 'ohai'
   });
 
-  assert.equal(get(service, 'queue.0.customProperty'), 'ohai');
+  assert.equal(get(service, 'queue.firstObject.customProperty'), 'ohai');
 });
 
 test('when preventDuplicates is `true`, setting a message is required', function(assert) {
--- a/tests/unit/services/flash-messages-test.js
+++ b/tests/unit/services/flash-messages-test.js
@@ -175,9 +175,14 @@ test('when preventDuplicates is `false` setting a message is not required', func
 test('when preventDuplicates is `true`, setting a message is required', function(assert) {
   set(service, 'defaultPreventDuplicates', true);
 
-  assert.throws((() => {
-    service.add({ });
-  }).message, 'Assertion Failed: The flash message cannot be empty when preventDuplicates is enabled.');
+  assert.throws(() => {
+      service.add({ });
+    },
+    ({ message }) => {
+      return message == 'Assertion Failed: The flash message cannot be empty when preventDuplicates is enabled.';
+    },
+    'Error is thrown'
+  );
 });
 
 test('it adds duplicate messages to the queue if preventDuplicates is `false`', function(assert) {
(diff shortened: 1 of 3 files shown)


tests/unit/services/flash-messages-test.js
```diff
@@ -162,6 +162,24 @@ test('it sets the correct defaults for service properties', function(assert) {
...
+  });
+
+  assert.equal(get(service, 'queue.0.customProperty'), 'ohai');
+});
+
+test('when preventDuplicates is `true`, setting a message is required', function(assert) {
+  set(service, 'defaultPreventDuplicates', true);
+
+  assert.throws((() => {
+    service.add({ });
```
sbatson5: I think `assert.throws` is the right solution to test this. However, you seem to be getting a false positive.  If you change this line to `service.add({ message: 'foo' })` your test will still pass. I think this is due to the wrapped function.

I think you will want something like this:

```
  assert.throws(() => {
    service.add({ });
  },
  ({ message }) => {
    return message == 'Assertion Failed: The flash message cannot be empty when preventDuplicates is enabled.';
  },
    'Error is thrown'
  );
```
`assert.throws` can take 3 arguments. If you pass it only 2, the second is simply the assertion description, so it was not matching the error text here.

The second argument should be a function that accepts the error object and returns true or false.  Give the above example a try and you should see it passes with no message and fails with one.

st-h: ah, thanks for catching that. Should all be fixed now :)

tests/unit/services/flash-messages-test.js
```diff
@@ -162,6 +162,29 @@ test('it sets the correct defaults for service properties', function(assert) {
...
 });
 
+test('when preventDuplicates is `false` setting a message is not required', function(assert) {
+  set(service, 'defaultPreventDuplicates', false);
+
+  service.add({
+    customProperty: 'ohai'
+  });
+
+  assert.equal(get(service, 'queue.0.customProperty'), 'ohai');
```
sbatson5: Minor nitpick, but rather than `queue.0` could we do `queue.firstObject`? I feel it reads just a bit more clear

st-h: yup, done :)

sbatson5:
Awesome! :shipit:

PR #263, opened 2018-02-16
update blueprint to handle newer test-helper.js

### LINKED ISSUE #262: Blueprint doesn't install test helper in Ember >= v2.17
I have some acceptance tests in my app which test for flash message content after an action.

After the v3 upgrade when these tests were converted to use `await`, then all started failing. It appears that `await` will wait for the ember-cli-flash timeout to end before moving on to the assertions that check for flash message content.

I created a demo app for this here: https://github.com/bgentry/ember3-flash-test-bug/blob/master/tests/acceptance/flash-text-test.js

Apologies if this turns out to be an Ember issue but I figured this is at least a good place to track it until that's sorted.

bgentry:
On newer Ember versions, test-helper.js does not include an `import { resolver }` line anymore. Instead, we insert after the last line in the default `test-helper.js`, the qunit import.

This should support [any version >= 2.17](https://github.com/ember-cli/ember-new-output/blame/v2.17.0/tests/test-helper.js#L3), which hopefully is good enough.

Fixes #262.

### DIFF
--- a/blueprints/ember-cli-flash/index.js
+++ b/blueprints/ember-cli-flash/index.js
@@ -1,4 +1,5 @@
 var EOL = require('os').EOL;
+let VersionChecker = require('ember-cli-version-checker');
 
 module.exports = {
   description: 'Generates ember-cli-flash test helper',
@@ -6,11 +7,20 @@ module.exports = {
   afterInstall: function() {
     var TEST_HELPER_PATH = 'tests/test-helper.js';
     var IMPORT_STATEMENT = EOL + "import './helpers/flash-message';";
-    var INSERT_AFTER = "import resolver from './helpers/resolver';";
+    var INSERT_AFTER_PRE_2_17 = "import resolver from './helpers/resolver';";
+    var INSERT_AFTER_POST_2_17 = "import { start } from 'ember-qunit';";
 
-    return this.insertIntoFile(TEST_HELPER_PATH, IMPORT_STATEMENT, {
-      after: INSERT_AFTER
-    });
+    let checker = new VersionChecker(this);
+    let ember = checker.forEmber();
+
+    let after;
+    if (ember.isAbove('2.17.0')) {
+      after = INSERT_AFTER_POST_2_17;
+    } else {
+      after = INSERT_AFTER_PRE_2_17;
+    }
+
+    return this.insertIntoFile(TEST_HELPER_PATH, IMPORT_STATEMENT, { after });
   },
 
   normalizeEntityName: function() {}
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "ember-cli-shims": "^1.1.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
+    "ember-cli-version-checker": "^2.1.0",
     "ember-disable-prototype-extensions": "^1.1.2",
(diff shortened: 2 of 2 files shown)


blueprints/ember-cli-flash/index.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   afterInstall: function() {
     var TEST_HELPER_PATH = 'tests/test-helper.js';
     var IMPORT_STATEMENT = EOL + "import './helpers/flash-message';";
-    var INSERT_AFTER = "import resolver from './helpers/resolver';";
+    var INSERT_AFTER = "import { start } from 'ember-qunit';";
```
sbatson5: Does this break on versions older than 2.17? Can we make this a ternary based on the version?

bgentry:
Is there a good way to check the version here? Couldn’t find any blueprints
that did that.

blueprints/ember-cli-flash/index.js
bgentry: It does only work on >=2.17:

> This should support any version >= 2.17, which hopefully is good enough.

FWIW, this addon has never installed correctly on those newer versions. But if there is an easy way to do an Ember version check within a blueprint, I agree it's worth doing.

sbatson5: We should be able to do it using `ember-cli-version-checker`:
https://github.com/ember-cli/ember-cli-version-checker

bgentry:
I updated this using ember-cli-version-checker to support versions either before v2.17 or after.

sbatson5:
Thanks for the update @bgentry. Would you mind using `yarn` and not committing the `package-lock` file? I personally have no preference, but the repo was switched to `yarn` in https://github.com/poteto/ember-cli-flash/pull/233

bgentry:
@sbatson5 dammit I have no idea how I committed that, sorry 😞 fixed now

package.json
```diff
@@ -48,6 +48,7 @@
     "ember-cli-shims": "^1.1.0",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^2.0.0",
+    "ember-cli-version-checker": "^2.1.0",
```
Dhaulagiri: i'm curious if this is going to work or not in consuming apps with this not being in `dependencies`

sbatson5: Just tested it on 2 apps.
Ember 2.8:
```
import resolver from './helpers/resolver';
import './helpers/flash-message';


import {
  setResolver
} from 'ember-qunit';

setResolver(resolver);
```
Ember 3:
```
import Application from '../app';
import config from '../config/environment';
import { setApplication } from '@ember/test-helpers';
import { start } from 'ember-qunit';
import './helpers/flash-message';


setApplication(Application.create(config.APP));

start();
```

Seems to be working

Dhaulagiri: I could definitely be wrong, but I wonder if this addon already exists in most scenarios given how common its usage is throughout the ember ecosystem.

Dhaulagiri: For instance yarn.lock didn’t need to be updated for it to be added here. We can always fix is someone notices anything not working.

sbatson5: haha, yeah: https://github.com/emberjs/ember.js/blob/master/package.json#L55

It's a dependency of `ember` so, I think we're getting it for free :P

Dhaulagiri: TIL!

PR #266, opened 2018-03-22
Upgrade to ember 3.0

jmar910:
Update dependencies and convert acceptance test to new unified testing goodness


sbatson5:
Thanks @jmar910. Did a quick glance and it looks good. I'll have some time tomorrow to look at why that one test is failing on `ember-canary`.

Dhaulagiri:
@sbatson5 I'm inclined to stop running tests against canary, what do you think?  The default travis setup ignores canary failures and I can't say I ever look at them for any addons, but I'm not sure how to do the same with circle without jumping through more hoops than seems worth it given that circle 1.0 is deprecated (i think).

sbatson5:
@Dhaulagiri I believe `ember-canary` is added to the default for `ember-try` but I believe it usually has an `allowedToFail` true. I'm fine updating that, so we can see the upstream failure but not block the PRs

Dhaulagiri:
Oh nice, I never realized you could do it right in ember-try itself 💯 

@jmar910 check out this [config](https://github.com/ember-cli/ember-try#configuration-files) which shows how we can add a `allowedToFail` property in the ember try config and move this along

jmar910:
Awesome! on it

Dhaulagiri:
turns out `allowedToFail` was already set for beta and canary @jmar910 but we removed it as we paired on this 🌵

jmar910:
lol yeah i noticed that whoops 😄

PR #269, opened 2018-06-06
add aria-label and others to flash message attrs

sbatson5:
## What's in this PR?

Ran into this problem on a client project where we needed to add some accessibility fields to these components.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -75,6 +75,7 @@ module.exports = function() {
         },
         {
           name: 'ember-beta',
+          allowedToFail: true,
           npm: {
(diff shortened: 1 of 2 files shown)


sbatson5:
@Dhaulagiri Think I could get a quick review here?

config/ember-try.js
```diff
@@ -75,6 +75,7 @@ module.exports = function() {
         },
         {
           name: 'ember-beta',
+          allowedToFail: true,
```
sbatson5: Beta should be allowed to fail, imo

PR #277, opened 2018-07-25
ember-cli 3.3

Dhaulagiri:


config/ember-try.js
```diff
@@ -22,25 +22,6 @@ module.exports = function() {
             }
           }
         },
-        {
-          name: 'ember-2.0',
```
sbatson5: Do we want this removed?

package.json
```diff
@@ -30,37 +25,37 @@
...
-    "ember-source": "~3.0.0",
+    "ember-source": "~3.3.0",
     "ember-source-channel-url": "^1.0.1",
     "ember-try": "^0.2.23",
     "eslint-plugin-ember": "^5.0.0",
-    "eslint-plugin-node": "^5.2.1",
-    "loader.js": "^4.2.3"
+    "eslint-plugin-node": "^6.0.1",
+    "loader.js": "^4.2.3",
+    "qunit-dom": "^0.6.2"
```
sbatson5: I know this is part of the default package now, but unless we start moving the tests over to use this, we can probably remove this dependency.  Maybe add it in a later PR?

config/ember-try.js
Dhaulagiri: my feeling has been that there isn't a lot of benefit to testing a release this old since there _shouldn't_ be much reason users would be stuck on 2.0 vs one of the later 2.x series releases.

package.json
Dhaulagiri: good point, i'll remove this

sbatson5:
Rebased off master

sbatson5:
🤔It looked like circle was only running on test scenario

sbatson5:
Addressing broken tests and config in https://github.com/poteto/ember-cli-flash/pull/285 as it was unrelated to this PR

sbatson5:
Rebased again

Issue #278, opened 2018-07-25
test suite no longer passes


Dhaulagiri:
@sbatson5 perhaps you might know better how to fix this, but due to a timing problem [this test] no longer passes in newer ember versions

PR #280, opened 2018-08-12
Use assign polyfill instead of Object.assign

### LINKED ISSUE #279: Fix for #274 (remove deprecated usages of Ember.copy) breaks ember-cli-flash in IE11 because of Object.assign usage
In the latest version 1.6.5 of ember-cli-flash we have two usages of `Object.assign` in services/flash-messages.js:

`const flashMessageOptions = Object.assign({}, defaults, { flashService });`

`const flashMessageOptions = Object.assign({}, options);`

`Object.assign` isn't supported in IE11, so version 1.6.5 of ember-cli-flash isn't working in IE11. To the best of my knowledge, IE11 should be supported by this add-on (as Ember officially supports it).

I suggest to use the Ember assign polyfill as described here:

https://www.emberjs.com/api/ember/3.3/functions/@ember%2Fpolyfills/assign

samcic:
Fixes #279 (IE11) issue with Object.assign

### DIFF
--- a/addon/services/flash-messages.js
+++ b/addon/services/flash-messages.js
@@ -13,6 +13,7 @@ import {
   equal
 } from '@ember/object/computed';
 import { classify } from '@ember/string';
+import { assign } from '@ember/polyfills';
 import { A as emberArray } from '@ember/array';
 import FlashMessage from 'ember-cli-flash/flash/object';
 import objectWithout from '../utils/object-without';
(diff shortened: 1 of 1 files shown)


samcic:
Note that the failing test here is not due to this PR: it seems the test has been failing for some time, see #278

Dhaulagiri:
This will be published in 1.6.6 as soon as I can figure out how to get 2fa and npm to work together nicely

Dhaulagiri:
Should be published now

PR #282, opened 2018-08-21
feat(service): handle flash object preventDuplicates similar to global option

### LINKED ISSUE #281: Allow setting preventDuplicates when adding a flash message
Hi,

I was wondering why `preventDuplicates` isn't really working but it turns out that the addon uses the globally configured `preventDuplicates`.

This means if one wants to have some toasts to prevent duplicates, but not all, it's currently not possible.
One could change the addon behavior to read `preventDuplicates` when adding a flash.

This means a new precedence would look like: 

1. flashMessages.add `preventDuplicates`
2. `defaultPreventDuplicates`

If this is ok, I could open a PR to implement this behavior

makepanic:
fixes #281

Everything open for discussion :)

### DIFF
--- a/addon/services/flash-messages.js
+++ b/addon/services/flash-messages.js
@@ -88,6 +88,8 @@ export default Service.extend({
   _newFlashMessage(options = {}) {
     assert('The flash message cannot be empty when preventDuplicates is enabled.',
         get(this, 'defaultPreventDuplicates') ? options.message : true);
+    assert('The flash message cannot be empty when preventDuplicates is enabled.',
+        get(options, 'preventDuplicates') ? options.message : true);
 
     const flashService = this;
     const allDefaults = getWithDefault(this, 'flashMessageDefaults', {});
@@ -155,7 +157,11 @@ export default Service.extend({
   },
 
   _enqueue(flashInstance) {
-    const preventDuplicates = get(this, 'defaultPreventDuplicates');
+    const instancePreventDuplicates = get(flashInstance, 'preventDuplicates');
+    const preventDuplicates = typeof instancePreventDuplicates === 'boolean' ?
+      // always prefer instance option over global option
+      instancePreventDuplicates :
+      get(this, 'defaultPreventDuplicates');
 
     if (preventDuplicates) {
       const guid = get(flashInstance, '_guid');
--- a/addon/utils/computed.js
+++ b/addon/utils/computed.js
@@ -2,6 +2,7 @@ import { typeOf } from '@ember/utils';
 import { computed, get } from '@ember/object';
 import { guidFor as emberGuidFor } from '@ember/object/internals';
 import { A as emberArray } from '@ember/array';
+import { isNone } from '@ember/utils';
 
 export function add(...dependentKeys) {
   const computedFunc = computed({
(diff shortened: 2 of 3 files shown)


addon/services/flash-messages.js
```diff
@@ -140,7 +142,8 @@ export default Service.extend({
   },
 
   _enqueue(flashInstance) {
-    const preventDuplicates = get(this, 'defaultPreventDuplicates');
+    const preventDuplicates = get(this, 'defaultPreventDuplicates') ||
+      get(flashInstance, 'preventDuplicates');
```
achambers: I'd have imagined that the option on the `flashInstance` itself should trump the global setting as it's more specific. This would not be the case with the above logic. What's the intention?

makepanic: You're right, that makes more sense.
I'll change it soon

makepanic:
I've made some changes so it'll always prefer the instance option over the global one.
Also rebased onto latest master

addon/utils/computed.js
```diff
@@ -30,6 +31,10 @@ export function guidFor(dependentKey) {
     get() {
       const value = get(this, dependentKey);
 
+      // it's possible that a value has no toString as some objects don't implement the guid field
+      // this early return it to avoid errors being thrown when calling undefined.toString()
+      if (isNone(value)) return;
```
alexparker: Perhaps just checking `if (typeof value.toString !== "function")` would show your intention more obviously?

makepanic: Sadly it would require additional checks as `typeof undefined.toString` will throw an error:

```
Uncaught TypeError: Cannot read property 'toString' of undefine
```

makepanic:
I've rebased this PR onto the current master

makepanic:
@sbatson5 @abhilashlr do you think we could also merge/work on this PR?

We used it since 2018 in production and haven't seen any issues.

If so, I'll check if I can find anything to improve. Right now it's only rebased onto the current master.

abhilashlr:
@sbatson5 Seems like the PR has test cases passing. If this is fine, I think we should merge it?

sarus:
Would really love to see this capability.

makepanic:
Rebased it onto the current master

tests/unit/services/flash-messages-test.js
```diff
@@ -205,6 +216,43 @@ module('FlashMessagesService', function(hooks) {
...
+    const messages = emberArray([ 'foo', 'foo', 'bar' ]);
+    const expectedResult = [ 'foo', 'foo', 'bar' ];
+    messages.forEach((message) => service.success(message, {preventDuplicates: false}));
+    const result = get(service, 'queue').mapBy('message');
+
+    assert.deepEqual(result, expectedResult, 'it adds duplicate messages to the queue');
+    assert.equal(get(service, 'queue').length, 3, 'it adds duplicate messages to the queue');
+  });
+
+  test('it does adds duplicate messages where preventDuplicates is `true`', function(assert) {
```
sbatson5, line 241: The test description here is a bit confusing to me. Aren't we saying it will honor the global default if an instance `preventDuplicates` is not set? i.e. we expect multiple `foo`'s but only one `bar` in this test case?

The test makes sense, I just find this wording a bit confusing

makepanic: Yes you're right.

I've adjusted the test title

PR #284, opened 2018-10-05
provide default config/overrides to FlashMessageService w/o injections

fivetanley:
While I was writing a component integration test in my app, I noticed
that I couldn't call `this.flashMessages.success` or any of the built in
convenience functions that target the bootstrap notifications.

This seems to be because, unless an application test runs first, the
service isn't getting set up with the right config. It looks like the
service is expecting it from only the initializer right now.

I changed app/services/flash-message.js to import the config from the
application, so that it is always set up correctly regardless of the
test context.

I also extracted the logic for merging the addon defaults and the app
overrides into a utility module so we could use the same logic in both
places as much as possible.

The unit test was updated to use the flash message service from the app instead (and not manually import the config). That was the best way I could think of to test it.

### DIFF
--- a/app/initializers/flash-messages.js
+++ b/app/initializers/flash-messages.js
@@ -2,41 +2,15 @@ import Ember from 'ember';
 import config from '../config/environment';
 /* eslint-disable ember/new-module-imports */
 const { deprecate } = Ember;
-const merge = Ember.assign || Ember.merge;
 const INJECTION_FACTORIES_DEPRECATION_MESSAGE = '[ember-cli-flash] Future versions of ember-cli-flash will no longer inject the service automatically. Instead, you should explicitly inject it into your Route, Controller or Component with `Ember.inject.service`.';
-const addonDefaults = {
-  timeout: 3000,
-  extendedTimeout: 0,
-  priority: 100,
-  sticky: false,
-  showProgress: false,
-  type: 'info',
-  types: [
-    'success',
-    'info',
-    'warning',
-    'danger',
-    'alert',
-    'secondary'
-  ],
-  injectionFactories: [
-    'route',
-    'controller',
-    'view',
-    'component'
-  ],
-  preventDuplicates: false
-};
+import flashMessageOptions from 'ember-cli-flash/utils/flash-message-options';
 
 export function initialize() {
   const application = arguments[1] || arguments[0];
   const { flashMessageDefaults } = config || {};
   const { injectionFactories } = flashMessageDefaults || [];
-  const options = merge(addonDefaults, flashMessageDefaults);
+  const options = flashMessageOptions(flashMessageDefaults);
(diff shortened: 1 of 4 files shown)


app/services/flash-messages.js
```diff
@@ -1 +1,12 @@
-export { default } from 'ember-cli-flash/services/flash-messages';
+import FlashMessagesService from 'ember-cli-flash/services/flash-messages';
+import { computed } from '@ember/object';
+import flashMessageOptions from 'ember-cli-flash/utils/flash-message-options';
+
+// import the application's config
+import config from '../config/environment';
```
Dhaulagiri: I wonder if we need to do something a little more robust like: `getOwner(this).resolveRegistration('config:environment');` to get the config?

fivetanley:
@Dhaulagiri thanks for the review! i updated based on the feedback, i didn't know you could resolve the config, that's neat.

to get that to work in the unit test, i had to use `setupTest` from ember test helpers/ember-qunit. Which means I also had to use the style of qunit where you put tests inside the module block, so that's why there's a lot of indentation noise there.

app/initializers/flash-messages.js
```diff
@@ -2,42 +2,16 @@ import Ember from 'ember';
...
 export function initialize() {
   const application = arguments[1] || arguments[0];
   const { flashMessageDefaults } = config || {};
   const { injectionFactories } = flashMessageDefaults || [];
-  const options = merge(addonDefaults, flashMessageDefaults);
+  const options = flashMessageOptions(flashMessageDefaults);
   const shouldShowDeprecation = !(injectionFactories && injectionFactories.length);
 
-  application.register('config:flash-messages', options, { instantiate: false });
-  application.inject('service:flash-messages', 'flashMessageDefaults', 'config:flash-messages');
```
Dhaulagiri: @fivetanley can you help me understand the change here? is this disabling the injection of the entire service, or just the config?

fivetanley: @Dhaulagiri just the config. i grep'd the code base to look for usage, and this was the only one I could find. I don't think it's necessary any longer because it was used to inject onto the service.

Breaking down these two lines a bit:

```javascript
application.register('config:flash-messages', options, { instantiate: false });
application.inject('service:flash-messages', 'flashMessageDefaults', 'config:flash-messages');
```
`application.register` _defines_ a singleton thing.
`application.inject` _injects_ a singleton thing onto another singleton thing.

I tend to think about it like this:

```javascript
application.register(singletonThing, options, { instantiate: false });
application.inject(targetToInjectPropertyOnto, propertyName, propertyValueThatIsTheResultOfResolving);
```

so the tl;dr of that is that this initializer was injecting a `flashMessageDefaults` property onto each instance of the `flash-messsages` service. Because we're looking up the config directly in `addon/services/flash-messages.js`, and it already has a `flashMessageDefaults`, we don't need to inject it here.

PR #286, opened 2018-10-12
add qunit-dom and update component tests

sbatson5:
## What's in this PR?

As discussed here: https://github.com/poteto/ember-cli-flash/pull/277#discussion_r206132419
We didn't want to add this dependency until it was actually used.  This PR updates the tests to use qunit-dom when applicable.  Also updates the modules for testing the component (integration and unit).

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "ember-try": "^0.2.23",
     "eslint-plugin-ember": "^5.0.0",
     "eslint-plugin-node": "^6.0.1",
-    "loader.js": "^4.2.3"
+    "loader.js": "^4.2.3",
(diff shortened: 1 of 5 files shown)


Dhaulagiri:
Looks like the tests are unhappy

sbatson5:
Yeah, seems to only be failing on certain ember version. I can investigate this weekend

Dhaulagiri:
@sbatson5 is this something you think you'll try to land at some point?

sbatson5:
@Dhaulagiri I've been meaning to come back to this, but work has been eating up too much time.  I have to figure out why this fails in one particular instance.

I can close for now to reduce repo noise

PR #303, opened 2019-09-27
fix: ensure timers are cleaned up on service destroy

makepanic:


makepanic:
I've noticed this because many of our acceptance tests lead into a flash appearing in the end (e.g. error flash or confirmation flash).

When investigating, there was always a dangling timer that slowed the test down:



With this PR:

sbatson5:
Awesome catch 👍 

I need to investigate the Circle failure -- it's unrelated to your change

abhilashlr:
@makepanic If you rebase with current master I believe it should pass this PR.

makepanic:
rebased it onto the current master

PR #304, opened 2019-10-26
remove deprecated mouse event handlers

### LINKED ISSUE #302: Using `mouseLeave` event handler methods in components has been deprecated
Using the latest Ember canary there is a new deprecation for `mouseLeave` and `mouseEnter`

https://github.com/poteto/ember-cli-flash/blob/09ffdfb0dc6fd8691d098b04c952ab955889e13a/addon/components/flash-message.js#L83-L95

```
bugsnag.js:662 DEPRECATION: Using `mouseLeave` event handler methods in components has been deprecated. [deprecation id: ember-views.event-dispatcher.mouseenter-leave-move] See https://emberjs.com/deprecations/v3.x#toc_component-mouseenter-leave-move for more details.
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:49220:21)
        at HANDLERS.<computed> (http://localhost:4200/assets/vendor.js:49317:9)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:49247:9)
        at HANDLERS.<computed> (http://localhost:4200/assets/vendor.js:49317:9)
        at invoke (http://localhost:4200/assets/vendor.js:49329:9)
        at deprecate (http://localhost:4200/assets/vendor.js:49285:28)
        at Class.init (http://localhost:4200/assets/vendor.js:16178:73)
        at Class.superWrapper (http://localhost:4200/assets/vendor.js:43644:22)
        at Class.init (http://localhost:4200/assets/vendor.js:80091:19)
```

These events should probably be migrated to use modifiers.

st-h:
fixes #302


sbatson5:
Thanks for doing this! Can you update the branch so we can get the test suite passing?

st-h:
merged in current master. looks good so far

addon/components/flash-message.js
```diff
@@ -58,6 +58,14 @@ export default Component.extend({
       set(this, 'active', true);
     });
     set(this, 'pendingSet', pendingSet);
+    this.element.addEventListener('mouseenter', this._mouseEnter);
```
Turbo87: are you sure that this actually works? it looks like `_mouseEnter()` is called without a `this` reference, so `const flash = get(this, 'flash');` would always return `undefined` 🤔

st-h: Good catch. I think you are right about this - looks like I forgot to add the action helper. However, I noticed doing so breaks the test (I think because they run on ember 3.8.3). So I used bind() to bind this to the functions to handle the events. Opened #315 to address this

PR #309, opened 2019-11-13
CHORE: Update node version on CircleCI to 8

abhilashlr:
This PR addresses #310

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -64,7 +64,6 @@ module.exports = function() {
         },
         {
           name: 'ember-release',
-          allowedToFail: true,
           npm: {
             devDependencies: {
               'ember-source': urls[0]
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -47,7 +47,24 @@ module.exports = function() {
           }
         },
         {
+          name: 'ember-lts-3.4',
+          npm: {
+            devDependencies: {
+              'ember-source': '~3.4.0'
+            }
+          }
+        },
+        {
+          name: 'ember-lts-3.8',
+          npm: {
+            devDependencies: {
+              'ember-source': '~3.8.0'
+            }
+          }
+        },
+        {
           name: 'ember-release',
+          allowedToFail: true,
           npm: {
             devDependencies: {
               'ember-source': urls[0]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ aliases:
 
   - defaults: &defaults
       docker:
-        - image: circleci/node:boron
+        - image: circleci/node:8
       working_directory: ~/code
 
   - &restore-repo-cache
@@ -20,6 +20,7 @@ aliases:
         - ~/.cache/yarn
   - &yarn
     |
+      node --version
       yarn install --non-interactive --cache-folder ~/.cache/yarn --frozen-lockfile
 
 jobs:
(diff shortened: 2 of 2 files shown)


config/ember-try.js
```diff
@@ -46,8 +46,25 @@ module.exports = function() {
...
+          name: 'ember-lts-3.8',
+          npm: {
+            devDependencies: {
+              'ember-source': '~3.8.0'
+            }
+          }
+        },
         {
           name: 'ember-release',
+          allowedToFail: true,
```
sbatson5: 🤔 Do we want to allow release to fail?  Beta and Canary make sense, but release feels important.

Looking through the logs, it seems to be getting tripped up on `lookup`
https://github.com/poteto/ember-cli-flash/blob/master/tests/helpers/utils/lookup.js#L6

makepanic: we might be able to use the [`getContext()` test helper](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#getcontext) until we refactor it to use `this.owner`

abhilashlr: 🤦‍♂ I was working on the version upgrade and added this props temporarily. Missed removing it. Will get that sorted👍

abhilashlr: It looks like the lookup util you had pointed @sbatson5 is not the problem. Migrating the integration test code to the latest `ember-qunit` format makes the tests to pass. However, this change is done in #308. Should I try merging this branch into the integration test branch and check if that works?

sbatson5: I'd definitely like to get the whole suite passing and supporting ember-release before we release a major version, but I'm happy to do this as a separate PR.  How much effort is it to get it passing with `ember-release` 😅 ?

sbatson5: Let's remove this flag and merge, even if the suite is failing (since we'll still be in the same state).  Then we can update https://github.com/poteto/ember-cli-flash/pull/308, since that isn't even running due to the node version

abhilashlr: Awesome. Sounds good👍. I'll remove the flag.

abhilashlr: @sbatson5 the flag is now removed

Turbo87:
the `package.json` file claims that the addon is compatible with Node 4 (see https://github.com/poteto/ember-cli-flash/blob/master/package.json#L18). shouldn't we run CI on Node 4 then? 🤔

abhilashlr:
> the `package.json` file claims that the addon is compatible with Node 4 (see https://github.com/poteto/ember-cli-flash/blob/master/package.json#L18). shouldn't we run CI on Node 4 then? 🤔

I will have this updated in `package.json` @Turbo87

abhilashlr:
> the `package.json` file claims that the addon is compatible with Node 4 (see https://github.com/poteto/ember-cli-flash/blob/master/package.json#L18). shouldn't we run CI on Node 4 then? 🤔

Also, this upgrade was done since Travis had been failing for sometime with PRs on this repo. Should we still support Node < 8?

Turbo87:
> Should we still support Node < 8?

No, dropping support for these outdated Node versions is fine, but it should be done in a major version release because it should be considered a breaking change.

abhilashlr:
> > Should we still support Node < 8?
> 
> No, dropping support for these outdated Node versions is fine, but it should be done in a major version release because it should be considered a breaking change.

Agree.

PR #315, opened 2019-12-01
fix missing bind on mouse events listener

st-h:
This fixes missing binds for mouse event listeners as noticed in #304 . Tried to use the action helper, but this seems to only be available in recent ember versions, therefore using `bind()` to address this


addon/components/flash-message.js
```diff
@@ -58,14 +58,16 @@ export default Component.extend({
       set(this, 'active', true);
     });
     set(this, 'pendingSet', pendingSet);
-    this.element.addEventListener('mouseenter', this._mouseEnter);
-    this.element.addEventListener('mouseleave', this._mouseLeave);
+    this.set('_mouseEnterHandler', this._mouseEnter.bind(this));
+    this.set('_mouseLeaveHandler', this._mouseLeave.bind(this));
```
abhilashlr: Should we approach with 1 format here instead of having a `this.set` and `set(this, ...)` mixed in code? In fact it would be nicer to use `Ember.setProperties(this, {})`. Thoughts? @Turbo87 @sbatson5

Turbo87: since `_mouseEnterHandler` is never observed in the template you can also just use `this._mouseEnter = this._mouseEnter.bind(this);` to simplify it a little bit

sbatson5: I'd lean toward `Ember.set` since that's the convention used elsewhere.

Turbo87:
@sbatson5 @st-h is there anything we can do to move this forward?

st-h:
@Turbo87 I am good with either approach, so I guess it's up to the maintainers do decide which to use

addon/components/flash-message.js
dcyriller: @sbatson5 I'm confused: you 👍ed @Turbo87's suggestion and proposed another alternative. Which one would you like to be implemented?

sbatson5: @dcyriller Sorry for the confusion.  I was 👍ing the suggestion that `this.foo = x` is possible rather than `this.set('foo', x)`.  However, given that that this repo is still using `Ember.set('foo', x)` everywhere, I would lean toward that approach.

Issue #320, opened 2020-02-14
Cannot read property 'removeEventListener' of null


maxymczech:
Hello. Lately I have been getting this error message from `ember-cli-flash` in Sentry console of a project, error comes from `addon-tree-output/ember-cli-flash/components/flash-message.js`:

```
    willDestroyElement() {
      this._super(...arguments);
      this.element.removeEventListener('mouseenter', this._mouseEnter);
      this.element.removeEventListener('mouseleave', this._mouseLeave);
    }
```

Is it possible to wrap this call in `if (this.element) { ... }`?

sbatson5:
Thank you for posting. This should be fixed now by https://github.com/poteto/ember-cli-flash/pull/315

PR #321, opened 2020-02-23
Fix event handler removal bug

maxymczech:
Fixes issue #320


sbatson5:
This was fixed in https://github.com/poteto/ember-cli-flash/pull/315

PR #326, opened 2020-08-24
Fix `getWithDefault` deprecation warning

sandydoo:
Starting with Ember v3.21, `getWithDefault` is deprecated and will be removed in v4.0.

https://emberjs.github.io/rfcs/0554-deprecate-getwithdefault.html

This PR adds the function as a utility, complete with tests.

I intentionally avoided the nullish operator `??` as it isn't available, based on the project's eslint config.
I did, however, add an explicit check for `null`. [Ember's default implementation does not do this](https://github.com/emberjs/ember.js/blob/93fb76515ec4b808962ea68ae95b11ead6986681/packages/%40ember/-internals/metal/lib/property_get.ts#L180), but it seems appropriate to add given the issues brought up in the RFC and provides a direct upgrade path to using `??` without `get`. I've checked all the usages of `getWithDefault` and believe this change shouldn't cause any unexpected issues.

### DIFF
--- a/addon/utils/get-with-default.js
+++ b/addon/utils/get-with-default.js
@@ -1,0 +1,11 @@
+import { get } from '@ember/object';
+
+export default function getWithDefault(objectInstance, key, defaultValue) {
+  let value = get(objectInstance, key);
+
+  if (value === undefined || value === null) {
+    return defaultValue;
+  }
+
+  return value;
+}
--- a/addon/utils/get-with-default.js
+++ b/addon/utils/get-with-default.js
@@ -1,9 +1,12 @@
 import { get } from '@ember/object';
 
+// This replicates Ember's deprecated `getWithDefault`.
+// Note that, as in the original, `null` is considered a valid value and will
+// not cause the function to return the default value.
 export default function getWithDefault(objectInstance, key, defaultValue) {
   let value = get(objectInstance, key);
 
-  if (value === undefined || value === null) {
+  if (value === undefined) {
     return defaultValue;
   }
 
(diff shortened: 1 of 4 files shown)


dbendaou (approved):
Seems to be the good way to go with this deprecation!

addon/utils/get-with-default.js
```diff
@@ -0,0 +1,11 @@
+import { get } from '@ember/object';
+
+export default function getWithDefault(objectInstance, key, defaultValue) {
+  let value = get(objectInstance, key);
+
+  if (value === undefined || value === null) {
```
sbatson5, line 6: I agree that a `null` check is clearer, but I feel like this could technically be a breaking change. If a user explicitly has `flash. destroyOnClick` set as `null`, Ember right now returns null.  This change would give them a default value instead.

Can we have it just mimic the current `getWithDefault` implementation?

sandydoo: > this could technically be a breaking change

Yeah, I've been thinking about this for the past few days. To be safe, this would have to be a major version bump, so perhaps it's just easier to stick with the same behaviour. I'll push an update.

chrismllr:
Looks like an intermittent failure on the ember-release scenario, might just be able to re-run for it to succeed

dbendaou:
Glad to see this merged :) Thanks!
Could we release a new version having this PR in it ?

sbatson5:
I'm getting ready for a 2.0.0 release but I will cut a 1.9.1 to backfill this deprecation fix

PR #328, opened 2020-09-29
Updated syntax to modern standards

jwlawrence:
- Converted to classes and decorators
- Replaced component lifecycle hooks with ember-render-modifiers
- Use this. rather than get
- Introduced some quality of life babel plugins
- Implemented angle bracket components


jwlawrence:
Before I look at the failures in older LTS versions, do we still want to support them?

addon/components/flash-message.js
```diff
@@ -2,127 +2,110 @@ import { htmlSafe, classify } from '@ember/string';
...
 
-  flashType: computed('flash.type', {
-    get() {
-      const flashType = getWithDefault(this, 'flash.type', '');
+const { next, cancel } = run;
 
-      return classify(flashType);
-    }
-  }),
+@tagName('')
```
luketheobscure, line 12: Is this just to make it easier to transition to glimmer components? Might be a breaking change for some consumers with little benefit.

addon/components/flash-message.js
```diff
@@ -2,127 +2,110 @@ import { htmlSafe, classify } from '@ember/string';
...
+  @and('showProgress', 'notExiting')
+  showProgressBar;
+
+  @readOnly('flash.exiting')
+  exiting;
+
+  @bool('template')
+  hasBlock;
+
+  @computed('flash.type')
```
luketheobscure, line 33: Missing a `messageStyle` dependency? Newer versions of eslint-plugin-ember can find and fix these kinds of errors.

addon/components/flash-message.js
```diff
@@ -2,127 +2,110 @@ import { htmlSafe, classify } from '@ember/string';
...
+  }
+
+  @action
+  onDidInsert(element) {
+    const pendingSet = next(this, () => {
+      set(this, 'active', true);
+    });
+    set(this, 'pendingSet', pendingSet);
+    set(this, '_mouseEnterHandler', this._mouseEnter.bind(this));
+    set(this, '_mouseLeaveHandler', this._mouseLeave.bind(this));
```
luketheobscure, line 99: Is there an advantage to doing this vs. using the `@action` decorator?

addon/components/flash-message.js
jwlawrence: Yeah, you're probably right, I didn't run try before I opened the PR so I didn't realize how many older versions broke. I would attempt to move this work to the 2.0 branch, but that hasn't been touched in four years. I was attempting to breathe some life into this project, but seems like without the help of the maintainers it will be an uphill battle. I'll close this PR.

addon/components/flash-message.js
jwlawrence: Not sure what you mean? I don't see the issue?

addon/components/flash-message.js
jwlawrence: This function does use the `@action` decorator and is called on "init" via the `did-insert` helper in the template.

addon/components/flash-message.js
luketheobscure: Sorry, I think this PR is 99% of the way there! Nearly all these changes are net wins. The only one I was worried about was this tagName decorator.

addon/components/flash-message.js
luketheobscure: `alertType` won't update when `this.messageStyle` changes. Just needs line 33 updated to `@computed('flash.type', 'messageStyle')`

addon/components/flash-message.js
luketheobscure: Should be able to get rid of the `.bind` stuff then. `@action` does this for you!

addon/components/flash-message.js
jwlawrence: Oh gotcha, I didn't look at the body heh. I'll update computed and see if I can get the linter to pick up this issue.

addon/components/flash-message.js
jwlawrence: Ah, I suppose I could leave this PR open and leave it up to the maintainers to determine if they're ok breaking older versions, I just assumed that it would be a deal-breaker.

As for your concerns about making this tagless, I'll point out that I added a wrapping div to the template that implements all the bindings we had before, namely classnames and applying various attributes, which is now handled via `...attributes`.

That said, this would introduce a change in markup in apps that do provide a custom tag type to the component and could potentially break them. I think going forward this approach is more in line with where we're heading with glimmer components, but I can remove this change if you think it's going to break consumers.

addon/components/flash-message.js
jwlawrence: Good call 👍

addon/components/flash-message.js
luketheobscure: I don't have any data to back it up, but I have a hunch that customizing the tagName is pretty rare. Adding the div in the template satisfies my main concern which was breaking styles that relied on that HTML structure.

addon/components/flash-message.js
```diff
@@ -2,127 +2,110 @@ import { htmlSafe, classify } from '@ember/string';
...
+  @action
+  onDidInsert(element) {
+    const pendingSet = next(this, () => {
+      set(this, 'active', true);
+    });
+    set(this, 'pendingSet', pendingSet);
+    set(this, '_mouseEnterHandler', this._mouseEnter);
+    set(this, '_mouseLeaveHandler', this._mouseLeave);
+    element.addEventListener('mouseenter', this._mouseEnterHandler);
+    element.addEventListener('mouseleave', this._mouseLeaveHandler);
```
luketheobscure, line 101: I'm not positive, but I think you might need the `@action` decorator on `this._mouseEnter` and `this._mouseLeave` to preserve the context. Other than that this all looks great!

luketheobscure (approved):

addon/components/flash-message.js
jwlawrence: Yeah, that was my thought too. If a user does use a custom tag it would change their markup only slightly, but it should be an easy fix (famous last words).

addon/components/flash-message.js
jwlawrence: Since the `@action` decorator is used on `onDidInsert` the references in its body will be bound to the component context, and since the handler functions are internal and not mutable (shouldn't be anyway) I don't think we should have an issue leaving them as "private" functions without the action decorator. I could be mistaken though, nothing is ever straightforward when dealing with `this`. The dummy app seems to behave as expected though and all tests pass.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-cli-flash",
-  "version": "1.9.0",
+  "version": "1.10.0",
```
sbatson5, line 3: Can we not include a version bump here?  Since we drop support for older versions of ember and require newer versions of node, I would consider this a major version.  I can cut that as a separate PR

sbatson5 (commented):
Thank you for taking the time to this 🙌 

Just one comment about the version number

PR #335, opened 2020-10-28
feat: added messageStylePrefix variable

Techn1x:
Enable the ability to not use the generic bootstrap or foundation class names, by adding ~~`flashTypePrefix`~~`messageStylePrefix` which provides a way for the user to specify the flash type class name prefix.

Backwards compatibility is maintained with `messageStyle`, but if ~~`flashTypePrefix`~~`messageStylePrefix` is set it will take priority.

```hbs
{{#each flashMessages.queue as |flash|}}
  <FlashMessage @flash={{this.flash}} @messageStylePrefix="special-alert-" />
{{/each}}
```
would produce an element like
```html
<div class="flash-message special-alert-success" > ... </div>
```

#### Rationale
I've been running a custom flash-message component for this addon for a while now, because;
* the flash-message component in this addon was a bit old
* the classes `alert` and `alert-box` are very overused and poorly scoped in our 15-20 frontend apps... we really needed to avoid using these class names, and this addon didn't provide a way to customize that (without resorting to a custom component)

I noticed that with v2.0.0 the component has been updated to be tagless, render modifiers etc so I thought I would contribute some of my workings back to the project, with the aim of being able to use the component in this addon again!

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,15 @@ By default, flash messages will have Bootstrap style class names. If you want to
 {{/each}}
 ```
 
+### Styling with user-specified flash type class prefix
+If you don't wish to use the class names associated with Bootstrap / Foundation, specify the `flashTypePrefix` on the component. This will override the class name prefixes with your own. For example, `flashTypePrefix='special-alert-'` would create flash messages with the class `special-alert-succcess`
+
+```handlebars
+{{#each flashMessages.queue as |flash|}}
+  {{flash-message flash=flash flashTypePrefix='special-alert-'}}
+{{/each}}
+```
+
 ### Sort messages by priority
 To display messages sorted by priority, add this to your template:
 
--- a/README.md
+++ b/README.md
@@ -377,12 +377,12 @@ By default, flash messages will have Bootstrap style class names. If you want to
 {{/each}}
 ```
 
-### Styling with user-specified flash type class prefix
-If you don't wish to use the class names associated with Bootstrap / Foundation, specify the `flashTypePrefix` on the component. This will override the class name prefixes with your own. For example, `flashTypePrefix='special-alert-'` would create flash messages with the class `special-alert-succcess`
+### Styling with user-specified message type class prefix
+If you don't wish to use the class names associated with Bootstrap / Foundation, specify the `messageStylePrefix` on the component. This will override the class name prefixes with your own. For example, `messageStylePrefix='special-alert-'` would create flash messages with the class `special-alert-succcess`
 
 ```handlebars
 {{#each flashMessages.queue as |flash|}}
-  {{flash-message flash=flash flashTypePrefix='special-alert-'}}
+  {{flash-message flash=flash messageStylePrefix='special-alert-'}}
 {{/each}}
 ```
 
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ If you don't wish to use the class names associated with Bootstrap / Foundation,
 
 ```handlebars
 {{#each flashMessages.queue as |flash|}}
-  {{flash-message flash=flash messageStylePrefix='special-alert-'}}
+  <FlashMessage @flash={{flash}} @messageStylePrefix='special-alert-' />
 {{/each}}
 ```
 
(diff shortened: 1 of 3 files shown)


Techn1x:
btw! I am participating in hacktoberfest this year so I would much appreciate it if this pull request could have the `hacktoberfest-accepted` label added to it, or if the repo could have `hacktoberfest` label/topic :100:

Techn1x:
Any interest in merging this @sbatson5 ? It's a requirement for me to not need to use a custom flash component

addon/components/flash-message.js
```diff
@@ -30,17 +30,20 @@ export default class FlashMessage extends Component {
...
-    const messageStyle = this.messageStyle || '';
-    let prefix = 'alert alert-';
 
-    if (messageStyle === 'foundation') {
-      prefix = 'alert-box ';
-    }
+    const { flashTypePrefix, messageStylePrefix } = this;
 
-    return `${prefix}${flashType}`;
+    const prefix = flashTypePrefix ? flashTypePrefix : messageStylePrefix;
```
sbatson5, line 45: Could this be `const prefix = flashTypePrefix || messageStylePrefix`?

README.md
```diff
@@ -377,6 +377,15 @@ By default, flash messages will have Bootstrap style class names. If you want to
 {{/each}}
 ```
 
+### Styling with user-specified flash type class prefix
+If you don't wish to use the class names associated with Bootstrap / Foundation, specify the `flashTypePrefix` on the component. This will override the class name prefixes with your own. For example, `flashTypePrefix='special-alert-'` would create flash messages with the class `special-alert-succcess`
```
sbatson5, line 381: Rather than calling it `flashTypePrefix` what are your thoughts on naming it `messageStylePrefix`? I feel like that would be clearer to the user that they are overriding `messageStyle`

addon/components/flash-message.js
```diff
@@ -30,17 +30,20 @@ export default class FlashMessage extends Component {
...
+  @computed('flash.type', 'flashTypePrefix', 'messageStylePrefix')
   get alertType() {
     const flashType = this.flash.type || '';
-    const messageStyle = this.messageStyle || '';
-    let prefix = 'alert alert-';
 
-    if (messageStyle === 'foundation') {
-      prefix = 'alert-box ';
-    }
+    const { flashTypePrefix, messageStylePrefix } = this;
```
sbatson5, line 43: What do you thin about adding a warning for the developer if they have both of these things set on the component? i.e. if they have `messageStyle` defined but also set `flashTypePrefix` they might have an expectation that both will be honored.  We could add a simple console warning saying that that `messageStyle` isn't being enforced

sbatson5 (changes requested):
Sorry it took me so long to get back to you. I thin this is a useful change. Just a couple comments around naming and quality of life

Techn1x:
Thanks @sbatson5 those are all great suggestions, implemented them all.

addon/components/flash-message.js
Techn1x: I did have this implemented, but then removed it, because `messageStyle` has a default value of `bootstrap` so it's always defined.

sbatson5: 🤦 Oh right

addon/components/flash-message.js
```diff
@@ -29,16 +29,16 @@ export default class FlashMessage extends Component {
...
   get alertType() {
     const flashType = this.flash.type || '';
-    const messageStyle = this.messageStyle || '';
-    let prefix = 'alert alert-';
-
-    if (messageStyle === 'foundation') {
-      prefix = 'alert-box ';
-    }
-
+    const prefix = this.messageStylePrefix || this._defaultMessageStylePrefix || '';
```
sbatson5, line 41: Is there a reason we have `|| ''`? Isn't it impossible for `_defaultMessageStylePrefix` to return a falsy value?

Techn1x: You're right, was just being overly careful

sbatson5 (approved):

PR #336, opened 2020-10-28
feat: improved a11y for flash message component by default

Techn1x:
Just noticed this while working on https://github.com/poteto/ember-cli-flash/pull/335

I _think_ `role="alert"` is what we want. There's also `role="alertdialog"` but I think that's more for when there is a decision to be made in a dialog context (eg yes/no/cancel)


sbatson5 (approved):
Good call. Thank you!

PR #339, opened 2020-11-03
update readme to modern standards

Techn1x:
angle bracket syntax for components, and native getters


sbatson5 (approved):
🙌 Thank you for this

PR #342, opened 2020-12-08
fix: rename flash Object class to FlashObject to avoid transpilation issues

### LINKED ISSUE #341: Throws when used with ember-cli-babel `includeExternalHelpers: true`
Hi,

when using ember-cli-flash in a project that enables [`includeExternalHelpers`](https://github.com/babel/ember-cli-babel#external-helpers) in its `ember-cli-babel` settings, the transpiled code will throw.

This happens because the `addon/flash/object` defines a `Object` class. During the transpilation babel adds the esModule marker:

```js
Object.defineProperty(_exports, "__esModule", {
    value: true
});

// ...

var Object = (_dec = (0, _computed.guidFor)('message').readOnly(), (_class = (_temp = /*#__PURE__*/function (_Ember$Object$extend) {
```

Due to variable hoisting, it'll try to call `defineProperty` on the class that ember-cli-flash defined.

If it's ok, I'll open a PR to rename that class name to something like `FlashObject` which prevents this from happening.

makepanic:
fixes #341

### DIFF
--- a/addon/flash/object.js
+++ b/addon/flash/object.js
@@ -3,7 +3,9 @@ import EmberObject, { set } from '@ember/object';
 import { cancel, later } from '@ember/runloop';
 import { guidFor } from '../utils/computed';
 
-export default class Object extends EmberObject.extend(Evented) {
+// Note:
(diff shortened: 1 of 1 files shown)


sbatson5:
Change makes sense to me 👍

Issue #354, opened 2021-04-19
Fix `this-property-fallback` deprecation warnings


aoumiri:
We would like to reduce the amount of deprecation warnings on our code base, and we noticed that we have some `this-property-fallback` related ones that are coming from this [template file](https://github.com/poteto/ember-cli-flash/blob/master/addon/templates/components/flash-message.hbs#L10).

A fix could be to just use `this.flash` instead of using it directly.

PR #355, opened 2021-04-19
Fix `this-property-fallback` deprecation warnings

aoumiri:
This MR fixes https://github.com/poteto/ember-cli-flash/issues/354


sbatson5 (approved):
Thanks for fixing this!

PR #373, opened 2021-12-28
Upgrade ember to v4

### LINKED ISSUE #370: Ember 3.26+: `manager-capabilities.modifiers-3-13` deprecation
Ember 3.26 introduced the [3.13 Modifier Manager Capabilities deprecation](https://deprecations.emberjs.com/v3.x/#toc_manager-capabilities-modifiers-3-13)

This addon won't be usable under Ember 4.x until this deprecation is addressed.

Todo this, the dependency on `@ember/render-modifiers` needs to be updated to 2.0+

stukalin:
Closes #371 
Closes #370 

@sbatson5 perhaps it makes sense to merge this one before releasing your #372 ? As you can see i've upgraded to the latest v4 but i kept all the ember-try scenarios starting from v3.12.


addon/utils/computed.js
```diff
@@ -1,33 +1,7 @@
-import { typeOf } from '@ember/utils';
 import { computed, get } from '@ember/object';
 import { guidFor as emberGuidFor } from '@ember/object/internals';
-import { A as emberArray } from '@ember/array';
 import { isNone } from '@ember/utils';
 
-export function add(...dependentKeys) {
```
stukalin, line 7: This computed wasn't used anymore

config/ember-try.js
```diff
@@ -58,19 +74,6 @@ module.exports = async function () {
           devDependencies: {},
         },
       },
-      {
-        name: 'ember-default-with-jquery',
```
stukalin, line 62: not sure if we needed this scenario anymore, but we don't use jquery in the addon and it caused issues for the v4

GreatWizard (approved):

bartocc:
I believe this PR should have triggered a major semver update

from https://github.com/ef4/ember-auto-import/blob/main/docs/upgrade-guide-2.0.md

> addons that upgrade to ember-auto-import >= 2 will only work in apps that have ember-auto-import >= 2, so they should do their own semver major releases when they upgrade

PR #379, opened 2022-01-09
Replace deprecated assign polyfill

sandydoo:
https://deprecations.emberjs.com/v4.x/#toc_ember-polyfills-deprecate-assign


bartocc:
We also depend on this addon and Ember V4 compat would be great.

Can we know what's needed to move this PR forward?

PR #382, opened 2022-06-15
Convert `flash-message` to Glimmer component

charlesfries:
- Install `@glimmer/component`,  `@glimmer/tracking`
- Convert `flash-message` component to Glimmer component
- Use co-located template


package.json
```diff
@@ -44,6 +44,8 @@
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.6.0",
     "@embroider/test-setup": "^0.48.1",
+    "@glimmer/component": "^1.1.2",
+    "@glimmer/tracking": "^1.1.2",
```
rwwagner90, line 48: Both of these should move to `dependencies` otherwise, apps without glimmer would not be able to use this addon

rwwagner90 (changes requested):
Thanks for taking on this work! The only real change I think we need is moving glimmer to dependencies.

rwwagner90:
@charlesfries I updated a lot of things in the repo. Could you please rebase?

rwwagner90:
@charlesfries looks like a template lint error. Not sure why it is showing up all the sudden. Can you add a `template-lint-disable` comment for that?

rwwagner90 (approved):

rwwagner90:
Thanks for the PR!

PR #383, opened 2022-06-16
ember-cli 4.4.0, remove auto service injection

rwwagner90:
This is not necessarily breaking, but because we are updating some things, dropping Ember version support, and removing the old initializers, I am tagging this as breaking just in case.

Closes #360
Closes #379
Closes #380


bartocc:
thx @rwwagner90 👍

PR #387, opened 2022-08-25
Fix message type

ardeliatay:


stukalin:
@ardeliatay well, it's been a couple of years already but do you want to wrap up your pr by any chance?

PR #390, opened 2023-02-06
Convert to typescript & octane

vlascik:
PR rewrites as much as possible to TS & Octane.

There are some TSC/ESLint errors due to FlashObject still extending EmberObject. Not sure what to do about these - if the object is rewritten as an ES6 class, subtle bugs are going to appear due to `init()` vs `constructor` timing, and the API will be changed (FlashObjects will have to be created with `new`, instead of `EmberObject.create()`.

I'm not entirely sure if users of this addon actually create FlashObjects on their own though - if not, and the only usage is through FlassMessages service, the change could be done and tests rewritten without much disruption to consumers.


RobbieTheWagner:
@vlascik thanks for taking on this work! It is long overdue. I would recommend ignoring lint issues for now. It would be good to get these tests running and see if everything is passing.

NullVoxPopuli:
Gonna close this since the package is now a v2 addon! thanks @vlascik your work inspired the work that did landed!

NullVoxPopuli:
> As far as I can tell the addon is still in JS.

Yes, but everything has type declarations now! 🎉

PR #391, opened 2023-10-07
  Convert to V2 addon

gilest:
Classic big bundle of deferred maintenance. I've tried to change as little as possible while getting a working build and passing tests.

Attempted to use community blueprints and tools with a particular view to ease of maintenance long-term.

- [x] Package manager -> pnpm
- [x] Project layout -> pnpm workspace (monorepo)
- [x] `ember-cli-flash` package updated by [V2 Addon blueprint](https://github.com/embroider-build/addon-blueprint)
- [x] `test-app` package updated by Ember CLI app blueprint
  - [x] I also took this to 4.12 to get relatively updated tooling
- [x] Node -> 16 (Although in v2 addons this isn't so important as they are prebuilt)
- [x] Render modifiers -> Local modifier
- [x] Peer deps on @ember/string and ember-modifier

Breaking 💥 

- Consuming apps now require `ember-auto-import` v2
- Node 16 (should we go 18 right away?)
- Dropped Ember "Classic" support
- Dropped Ember 3.24 support. Minimum supported version is now 3.28

Unresolved/maybe breaking

- Type definitions – I've manually placed these in `declarations/` and tested this locally
- Releasing will be similar but the RELEASING file will need to be adjusted slightly

Todo soon after

- Migrate to typescript so that declarations can be built automatically with source mapping etc
- Replace and document the automatic (via CLI blueprint) test helper (I've extracted a replacement for this into its own branch, I'll open a PR)


.github/workflows/push-dist.yml
```diff
@@ -0,0 +1,25 @@
...
+      - main
+      - master
+
+jobs:
+  push-dist:
+    name: Push dist
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: NullVoxPopuli/action-setup-pnpm@v2
```
NullVoxPopuli, line 20:
```suggestion
      - uses: wyvox/action-setup-pnpm@v3
```

ember-cli-flash/package.json
```diff
@@ -0,0 +1,111 @@
...
+    },
+    {
+      "name": "Scott Batson"
+    },
+    {
+      "name": "Brian Runnells"
+    }
+  ],
+  "exports": {
+    ".": "./dist/index.js",
```
NullVoxPopuli, line 33: This file should probably have a types entry as well

ember-cli-flash/package.json
```diff
@@ -0,0 +1,111 @@
...
+    "eslint-plugin-ember": "^11.11.1",
+    "eslint-plugin-n": "^16.0.0",
+    "eslint-plugin-prettier": "^5.0.0",
+    "prettier": "^3.0.3",
+    "prettier-plugin-ember-template-tag": "^1.1.0",
+    "rollup": "^3.21.8",
+    "rollup-plugin-copy": "^3.4.0"
+  },
+  "engines": {
+    "node": "12.* || 14.* || >= 16"
```
NullVoxPopuli, line 79: This whole node block is not needed

ember-cli-flash/package.json
NullVoxPopuli: And maybe reexport to component and service type if that's not happening already

test-app/ember-cli-build.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = function (defaults) {
+  let app = new EmberApp(defaults, {
+    autoImport: {
+      watchDependencies: ['ember-cli-flash'],
+    },
+  });
+
+  const { maybeEmbroider } = require('@embroider/test-setup');
+  return maybeEmbroider(app);
```
NullVoxPopuli, line 13: Instead of maybe embroider, we courd do always strictest embroider. 
This would guarantee no accidental regressions are possible, and also provide a better debugging experience

NullVoxPopuli (changes requested):
Fantastic start! I'd nay this is very close!

Some comments can be in followup PRs, if that's easier

.github/workflows/push-dist.yml
gilest: Fair enough. Suggest you [update this in the addon blueprint](https://github.com/search?q=repo%3Aembroider-build%2Faddon-blueprint%20NullVoxPopuli%2Faction-setup-pnpm&type=code)

ember-cli-flash/package.json
```diff
@@ -0,0 +1,111 @@
...
+    "lint:hbs": "ember-template-lint . --no-error-on-unmatched-pattern",
+    "lint:hbs:fix": "ember-template-lint . --fix --no-error-on-unmatched-pattern",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "prepack": "rollup --config",
+    "start": "rollup --config --watch",
+    "test": "echo 'A v2 addon does not have tests, run tests in test-app'"
+  },
+  "dependencies": {
+    "@ember/render-modifiers": "^2.0.2",
```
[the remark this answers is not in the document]
gilest: Yeah shouldn't be too difficult but not doing it in this PR

test-app/ember-cli-build.js
gilest: Good idea

.github/workflows/push-dist.yml
NullVoxPopuli: yes!

ember-cli-flash/package.json
gilest: An index file doesn't actually exist in `/dist` (yet) so I've removed. We can add a top level module export with types as an enhancement later

ember-cli-flash/package.json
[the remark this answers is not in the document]
gilest: Ended up resolving as this package had peerDep issues

ember-cli-flash/package.json
```diff
@@ -0,0 +1,110 @@
...
+    "eslint-plugin-n": "^16.0.0",
+    "eslint-plugin-prettier": "^5.0.0",
+    "prettier": "^3.0.3",
+    "prettier-plugin-ember-template-tag": "^1.1.0",
+    "rollup": "^3.21.8",
+    "rollup-plugin-copy": "^3.4.0"
+  },
+  "peerDependencies": {
+    "@ember/string": "^3.1.1",
+    "ember-modifier": "^4.1.0"
```
gilest, line 78: Thoughts on these version strings? Could maybe be more permissive here with `>=`.

I don't want to support `ember-modifier` v3 because because of the awkward `eager: false` API.

NullVoxPopuli: yeah that's fair -- I think `>=` should probably be used as it's been a pain to move some ecosystem stuff from `^` to `>=` when the next major of whatever-it-is isn't actually breaking for whatever-the-package-is.

test-app/ember-cli-build.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = function (defaults) {
+  let app = new EmberApp(defaults, {
+    autoImport: {
+      watchDependencies: ['ember-cli-flash'],
+    },
+  });
+
+  const { Webpack } = require('@embroider/webpack');
+  return require('@embroider/compat').compatBuild(app, Webpack, {
+    staticAddonTestSupportTrees: true,
```
NullVoxPopuli, line 14: noice!

test-app/config/ember-try.js
```diff
@@ -56,14 +64,6 @@ module.exports = async function () {
...
-        npm: {
-          devDependencies: {
-            'ember-source': '~3.28.0',
-          },
-          ember: {
-            edition: 'classic',
-          },
-        },
       },
       embroiderSafe(),
```
NullVoxPopuli, line 68: we no longer need both of these, since we only test with embroider-strict

NullVoxPopuli, line 68:
```suggestion
```

test-app/config/ember-try.js
```diff
@@ -56,14 +64,6 @@ module.exports = async function () {
...
       embroiderOptimized(),
```
NullVoxPopuli, line 69:
```suggestion
```

gilest: Yeah of course

ember-cli-flash/package.json
NullVoxPopuli: _however_, I don't think either of these dependencies will change

NullVoxPopuli (approved):
I think this is good -- I'd like to get one other adopted-ember-addons maintainer to review tho -- just for knowledge spread.

This PR is pretty big, but I'd prefer we make progress and iterate in subsequent smaller PRs, if we need to

ember-cli-flash/package.json
gilest: Agreed

PR #392, opened 2023-10-23
Replace blueprint with explicit test helpers

gilest:
### Old world 🌐 

Previously the following was added to an Ember CLI app when installed via `ember install ember-cli-flash`.

```js
// tests/helpers/flash-message.js
import FlashObject from 'ember-cli-flash/flash/object';

FlashObject.reopen({ init() {} });
```

```js
// tests/test-helper.js
import './helpers/flash-message';
```

Essentially importing this helper would cause the `init` hook of the `FlashObject` to become a no-op and disable the timeout running in consuming apps test runs.

It wasn't very explicit or obvious that what was happening as a result of this import. Looking at apps that I maintain, some do and some don't have this...

### New world 🌏 

Blueprints are not supported in V2 addons, so this PR ships two new test helpers instead.

```js
import { disableTimeout, enableTimeout } from 'ember-cli-flash/test-support';
```

To get the same behaviour as "Old world 🌐 " a consuming app needs only to remove the blueprint import. ✅ 

However this can be changed at any time with the `enableTimers` function. This allows consuming apps to choose on a per-test basis whether Flash timeouts should occur – as demonstrated by the included tests.

### Why not disable timeouts for all test runs? 🤔 

There are many existing apps with test suites _**that both do and do not have timeouts disabled**_ at present I think it's important that this is configurable for compatibility reasons.

Even this addon's own test suite tests with the timeout functionality enabled.

### Other options 🤷🏻 

Very open to suggestions here. Just wanted to have a known way forward to replace the existing blueprint.

### Documentation

- [x] Added to README test helpers ([Rendered](https://github.com/adopted-ember-addons/ember-cli-flash/blob/3ebd637a4222631426219e1048f40a7c61aca353/README.md#test-helpers))
- [x] Added to UPGRADING ([Rendered](https://github.com/adopted-ember-addons/ember-cli-flash/blob/3ebd637a4222631426219e1048f40a7c61aca353/UPGRADING.md))


test-app/tests/unit/flash/object-test.js
```diff
@@ -2,10 +2,56 @@ import { run, later } from '@ember/runloop';
...
 import FlashMessage from 'ember-cli-flash/flash/object';
+import { disableTimeout, enableTimeout } from 'ember-cli-flash/test-support';
 
 const testTimerDuration = 50;
 let flash = null;
 
+module('FlashMessageObject disableTimeout', function (hooks) {
+  hooks.beforeEach(function () {
+    disableTimeout();
+    flash = FlashMessage.create({
```
NullVoxPopuli, line 13: are you planning to remove the `EmberObject` usage before v5?

https://github.com/adopted-ember-addons/ember-cli-flash/blob/master/ember-cli-flash/src/flash/object.js#L2

NullVoxPopuli (approved):

gilest:
Self-reivew: I think we can reduce the configuration a bit here by disabling timers by default.

test-app/tests/unit/flash/object-test.js
gilest: I wasn't specifically, although that will be a breaking change at some point...

ember-cli-flash/package.json
```diff
@@ -49,7 +49,8 @@
     "test": "echo 'A v2 addon does not have tests, run tests in test-app'"
   },
   "dependencies": {
-    "@embroider/addon-shim": "^1.0.0"
+    "@embroider/addon-shim": "^1.0.0",
+    "@embroider/macros": "^1.13.2"
```
gilest, line 53: I'm not sure this should be a peer since it's not in the app blueprint.

I get confused about some of these build packages, though 😅  Are we missing other dependencies here like `ember-auto-import`?

test-app/tests/test-helper.js
```diff
@@ -4,6 +4,9 @@ import * as QUnit from 'qunit';
 import { setApplication } from '@ember/test-helpers';
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
+import { enableTimeout } from 'ember-cli-flash/test-support';
+
+enableTimeout();
```
gilest, line 9: The test app needs to dogfood the `enableTimeout` helper since it tests timeout functionality of the addon itself.

gilest:
@NullVoxPopuli please add `breaking`

ember-cli-flash/package.json
NullVoxPopuli: Auto-import is only needed for v1 addons.

I think peer is correct, because it communicates that the consumer needs to do something

gilest: New PR for this #395

PR #394, opened 2023-10-23
Convert `FlashObject` to native class

gilest:
This will be `breaking` due to

- Public type declarations
- The `getFlashObject` [public API](https://github.com/adopted-ember-addons/ember-cli-flash?tab=readme-ov-file#returning-flash-object)

To do
- [x] Upgrade guide -> You cannot call `.get()` etc on FlashObject instances
- [x] Update type declarations

As for these two I'm not sure what their purpose is, I couldn't find any mention of them in the docs or PRs/issues. I don't think they are public API...

> - Document `on('didDestroyMessage')` event -> `onDidDestroyMessage` callback
> - Document `on('didExitMessage')` event -> `onDidExitMessage` callback


ember-cli-flash/src/flash/object.js
```diff
@@ -1,22 +1,22 @@
...
+    return guidFor(this.message?.toString());
+  }
 
-  // eslint-disable-next-line ember/classic-decorator-hooks
-  init() {
-    super.init(...arguments);
+  constructor(messageOptions) {
+    for (const [key, value] of Object.entries(messageOptions)) {
+      this[key] = value;
+    }
```
gilest, line 19: Should possibly only assign allow-listed keys 🤷🏻 or even enumerate the specific keys to assign 🤷🏻

ember-cli-flash/src/flash/object.js
```diff
@@ -40,25 +40,24 @@ export default class FlashObject extends EmberObject.extend(Evented) {
...
+  destroy() {
     if (this.onDestroy) {
       this.onDestroy();
     }
 
     this._cancelTimer();
     this._cancelTimer('exitTaskInstance');
-    super.willDestroy(...arguments);
+    this.isDestroyed = true;
   }
```
gilest, line 53: Renamed `willDestroy` hook to `destroy` so this object is sort of pretending to be an Ember object still (the tests check `isDestroyed`.

Not sure if there's operations that happen in the real `destroy()` that we need to handle here.

ember-cli-flash/src/flash/object.js
```diff
@@ -1,22 +1,22 @@
...
+import { tracked } from '@glimmer/tracking';
+import { guidFor } from '@ember/object/internals';
 
-// Note:
-// To avoid https://github.com/adopted-ember-addons/ember-cli-flash/issues/341 from happening, this class can't simply be called Object
-export default class FlashObject extends EmberObject.extend(Evented) {
+export default class FlashObject {
   exitTimer = null;
-  exiting = false;
+  @tracked exiting = false;
```
gilest, line 7: Tracked since this is the only value I'm aware of that can cause a render.

ember-cli-flash/src/flash/object.js
NullVoxPopuli: > ❓ Not sure if there's operations that happen in the real destroy() that we need to handle here.

nah, with `@ember/destroyable` the destroy callback can be a standalone function from anywhere

ember-cli-flash/src/flash/object.js
NullVoxPopuli: we'll probably want to disallow this at some point
because we can pass components as values (and should) and migrate away from string-based component references.

gilest: Ah – we should update the example, but that's just demonstrating that _any_ arbitrary keys can be set to any value.

UPGRADING.md
```diff
@@ -1,5 +1,15 @@
 # Upgrading ember-cli-flash
 
+## Upgrading to v6
```
gilest, line 3: Assuming next release will be v6, although minor, this is a breaking change.

Would be good to batch all the remaining internal changes and TypeScript migration in the same update...

ember-cli-flash/src/flash/object.js
```diff
@@ -1,34 +1,35 @@
...
 const defaultDisableTimeout = macroCondition(isTesting()) ? true : false;
 
-// Note:
-// To avoid https://github.com/adopted-ember-addons/ember-cli-flash/issues/341 from happening, this class can't simply be called Object
-export default class FlashObject extends EmberObject.extend(Evented) {
+export default class FlashObject {
   exitTimer = null;
-  exiting = false;
+  @tracked exiting = false;
+  @tracked message = '';
```
gilest, line 12: This this is no longer an EmberObject. Only the tracked properties will update rendered flash messages. We may want to track `type` as well, although I'm not sure if that would even work.

ember-cli-flash/src/flash/object.js
```diff
@@ -79,27 +80,27 @@ export default class FlashObject extends EmberObject.extend(Evented) {
     const timerTaskInstance = later(() => {
       this.exitMessage();
     }, this.timeout);
-    set(this, 'timerTaskInstance', timerTaskInstance);
+    this.timerTaskInstance = timerTaskInstance;
   }
 
   exitTimerTask() {
     if (this.isDestroyed) {
```
NullVoxPopuli, line 87: could also do `isDestroyed(this)` using `@ember/destroyable`

test-app/tests/integration/components/flash-message-test.js
```diff
@@ -18,23 +19,29 @@ module('Integration | Component | flash message', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it renders a flash message', async function (assert) {
-    this.set('flash', FlashMessage.create({ message: 'hi', sticky: true }));
+    const flash = new FlashMessage({ message: 'hi', sticky: true });
```
NullVoxPopuli, line 22: does the FlashMessage need `associateDestroyableChild` when it's created?

gilest: Yes, good idea. I've implemented this in `FlashMessagesService` when the messages are created

gilest (commented):
@NullVoxPopuli migrated the destruction to use `@ember/destroyable`. Seems to work good, there are quite a lot of tests that were checking destruction timing

ember-cli-flash/declarations/flash/object.d.ts
```diff
@@ -11,6 +11,5 @@ declare module 'ember-cli-flash/flash/object' {
     allowExit(): void;
     timerTask(): void;
     exitTimerTask(): void;
-    destroy(): void;
```
gilest: This was an `EmberObject` method. Public API `destroyMessage` will destroy the message and should be used instead.

NullVoxPopuli (approved):

PR #399, opened 2023-12-01
fix(types): add typesVersions map

simonihmig:
To allow consumers who are still on `"moduleResolution": "node"` in their TS config (i.e. where TS does not use `exports`) to resolve the types correctly.


ArnaudWeyts:
@gilest This would also fix an issue we are running into, and shouldn't result in a breaking change I suppose, are there any plans to merge this?

ArnaudWeyts (approved):

PR #400, opened 2024-02-01
Setup relase plan

NullVoxPopuli:
Sets up: https://github.com/embroider-build/release-plan

The more-automated replacement for release-it, allowing releases from the couch, car, you name it.

Changes:
1. `❯ npx create-release-plan-setup@latest --update`
2. remove release-it.


gilest (approved):
Sweet. Looks like the PR labels match what was being used. Let's see what it generates...

PR #409, opened 2024-03-12
Modernise `FlashMessagesService`

gilest:
Replaces computeds, EmberArray, set/get, prototype extensions with native JS features.

Should make the conversion to TypeScript much easier.

Breaking since the types of these properties will be changing. Targeting `6.0.0` with all of the other breaking modernisation changes. Hopefully we can ship new types in the same release.


ember-cli-flash/src/services/flash-messages.js
```diff
@@ -1,44 +1,42 @@
...
+import { tracked } from '@glimmer/tracking';
+import { registerDestructor } from '@ember/destroyable';
 
 export default class FlashMessagesService extends Service {
-  @(equal('queue.length', 0).readOnly())
-  isEmpty;
+  @tracked queue = [];
+
+  get arrangedQueue() {
+    return this.queue.sort(function (a, b) {
```
NullVoxPopuli, line 17: is it ok to use sort here since sort mutates?

ember-cli-flash/src/services/flash-messages.js
```diff
@@ -48,30 +46,28 @@ export default class FlashMessagesService extends Service {
...
   }
 
   peekFirst() {
-    return this.queue.firstObject;
+    return this.queue.at(0);
   }
 
   peekLast() {
-    return this.queue.lastObject;
+    return this.queue.at(-1);
```
NullVoxPopuli, line 70: Just in case someone has to support old Safari: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility

gilest: Interesting, I was wondering about this

 In a v1 addon would this have been transpiled by babel?

NullVoxPopuli: only if configured to use some polyfill

ember-cli-flash/src/services/flash-messages.js
gilest: Good point

gilest:
@NullVoxPopuli I have [pushed the upstream branch](https://github.com/adopted-ember-addons/ember-cli-flash/compare/master...gilest:ember-cli-flash:refactor/modernise-service) but GitHub hasn't updated the PR page for some reason. I'll circle back in a while and re-stage if it doesn't update

NullVoxPopuli (approved):

PR #415, opened 2025-03-04
Upgrade test-app to v5 app blueprint

gilest:
The main branch CI is failing due to ember-try scenarios for Ember v6.

Being a little conservative I've just taken the test-app to v5.

I think this is the simplest way to get the CI green again.

Will add comments to changes that differ from the blueprint.

Suggest label: `internal`


.github/workflows/ci.yml
```diff
@@ -48,8 +48,7 @@ jobs:
       matrix:
         try-scenario:
           - ember-lts-3.28
-          - ember-lts-4.4
-          - ember-lts-4.8
+          - ember-lts-4.12
```
gilest, line 51: There's some annoying resolution issues with < 4.11. Makes the most sense to test against 4.12 anyway.

test-app/config/ember-try.js
```diff
@@ -10,23 +10,19 @@ module.exports = async function () {
...
-      {
-        name: 'ember-lts-4.8',
-        npm: {
-          devDependencies: {
-            'ember-source': '~4.8.0',
+            'ember-cli': '~4.12.2',
+            'ember-source': '~4.12.0',
           },
         },
       },
```
gilest, line 28: Have to do a little dependency shuffling to get these old ember-source versions to still pass in CI. Might be time to drop some support soon 🤷🏻

test-app/.eslintrc.js
```diff
@@ -25,6 +25,8 @@ module.exports = {
     'ember/no-classic-classes': 'warn',
     'ember/no-computed-properties-in-native-classes': 'warn',
     'ember/classic-decorator-hooks': 'warn',
+    // TODO
+    'ember/no-runloop': 'warn',
```
gilest, line 29: Set these to warn since adressing them is out of scope of this PR

gilest:
@NullVoxPopuli why is this breaking 🤔

test-app/package.json
```diff
@@ -12,72 +12,73 @@
...
+    "stylelint-config-standard": "^34.0.0",
+    "stylelint-prettier": "^4.1.0",
+    "webpack": "^5.95.0"
   },
   "peerDependencies": {
     "ember-source": "^4.0.0"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
+    "node": ">= 18"
```
Windvis, line 81: I guess this could be considered a breaking change? I'm not sure which version this addon officially supported before, but CI seems to run against node 16.

Windvis (approved):
Looks good, and I don't think it's a breaking change either :+1:

